Issue #50, opened 2021-10-29
MQTT Bridge: connect to a broker in the same local network.


MarcoRBosco:
**Feature Description**
Allow to bridge broker in the same local network using Greengrass.

**Use Case**
Bridge broker in the local network of Greengrass to the cloud without giving certificates to it. This is required when the broker is already implemented and working.

**Proposed Solution**
![P21019_MQTT_Bridge](https://user-images.githubusercontent.com/62314876/139452243-616be18e-b553-4fca-a314-79f3d8c72404.png)

Use same configuration file as the component example [MQTT Bridge](https://github.com/aws-greengrass/aws-greengrass-component-examples/tree/65c420d6231d2f36999faec75636d05655dbcc15/mqtt-bridge):

```yaml
mappings:
    localTopics:
      - "/localTopic"
    cloudTopics:
      - "/cloudTopic"
connectionInfo:
    brokerUri: "tcp://localhost:1883"
    clientId: "bridger"
    username: null
    password: null
```

**Other**
NA

* [ ] :wave: I may be able to implement this feature request
* [ ] :warning: This feature might incur a breaking change

MarcoRBosco:
Hello Joe,

To put on context the feature request, we are working in a proof of concept where the first target is to validate AWS IoT Services where the architecture is as follows (ip are just example):

![P21019_MQTT_Bridge (1)](https://user-images.githubusercontent.com/62314876/139578405-b18e9720-e269-4c81-aa00-762a25ddcb19.png)

The instance with the third party software and the mosquitto broker is an old system set up by the end customer. The main idea after AWS IoT Services are validated is to remove instances with third party software and the mosquitto broker with AWS Greengrass V2. PLCs will be connected directly to Greengrass using Modbus or OPC-UA, but for this first step we need to bridge the unencrypted topics on port 1883 from the third party software. 

Hope this helps to understand a little bit the use case that motivates this feature request.

We are sorry but we are going out of time to purpose a PR we will use the [MQTT Bridge](https://github.com/aws-greengrass/aws-greengrass-component-examples/tree/65c420d6231d2f36999faec75636d05655dbcc15/mqtt-bridge).

Nevertheless thank you for accepting the feature request for an open discussion.

Thanks,
Marco

jbutler:
[...]

PR #55, opened 2022-03-09
feat: ignore CDA when brokerUri not ssl

jcosentino11:
**Issue #, if available:**
Partially addresses #50 

**Description of changes:**
When `brokerUri` uses an unencrypted protocol, integration with client-device-auth component is disabled

To avoid listening to cert updates from CDA when ssl is not being used for the broker connection. Supports use case where customers may want to connect to broker over tcp.

- [x] `mvn clean verify`
- [ ] UAT (still working on this)

### DIFF
--- a/src/main/java/com/aws/greengrass/mqttbridge/clients/MQTTClient.java
+++ b/src/main/java/com/aws/greengrass/mqttbridge/clients/MQTTClient.java
@@ -266,7 +266,12 @@ public class MQTTClient implements MessageClient {
     private void reconnectAndResubscribe() {
         int waitBeforeRetry = MIN_WAIT_RETRY_IN_SECONDS;
 
-        while (!mqttClientInternal.isConnected() && !Thread.interrupted()) {
+        while (!mqttClientInternal.isConnected()) {
+            // prevent connection attempt when interrupted
+            if (Thread.interrupted()) {
+                return;
+            }
+
             try {
                 // TODO: Clean up this loop
                 doConnect();


src/main/java/com/aws/greengrass/mqttbridge/MQTTBridge.java
```diff
@@ -118,29 +120,37 @@ public void install() {
 
     @Override
     public void startup() {
+        String serverUri = Coerce.toString(config.findOrDefault(MQTTClient.DEFAULT_BROKER_URI,
+                KernelConfigResolver.CONFIGURATION_CONFIG_KEY, MQTTClient.BROKER_URI_KEY));
+        boolean ssl = serverUri.startsWith("ssl");
```
MikeDombo, line 125: Parse into a URI and then check the protocol field.

src/main/java/com/aws/greengrass/mqttbridge/MQTTBridge.java
```diff
@@ -118,29 +120,37 @@ public void install() {
 
     @Override
     public void startup() {
+        String serverUri = Coerce.toString(config.findOrDefault(MQTTClient.DEFAULT_BROKER_URI,
```
MikeDombo, line 123: You'll need to subscribe to this topic so that if it changes in a deployment, the bridge will reconfigure itself. We could also setup bridge to just restart itself if any of these important fields change.

src/main/java/com/aws/greengrass/mqttbridge/MQTTBridge.java
```diff
@@ -118,29 +120,37 @@ public void install() {
...
-                } catch (IOException | CertificateException | KeyStoreException e) {
-                    logger.atError("Invalid CA list").kv("CAList", Coerce.toString(newv)).log();
-                    serviceErrored(String.format("Invalid CA list. %s", e.getMessage()));
-                }
-            });
+            if (ssl) {
+                kernel.locate(ClientDevicesAuthService.CLIENT_DEVICES_AUTH_SERVICE_NAME).getConfig()
+                        .lookup(RUNTIME_STORE_NAMESPACE_TOPIC, ClientDevicesAuthService.CERTIFICATES_KEY,
+                                ClientDevicesAuthService.AUTHORITIES_TOPIC).subscribe((why, newv) -> {
+                            try {
```
MikeDombo, line 141: I believe the on change callback here will break things if the bridge is configured with ssl, then is changed to be without ssl.

src/main/java/com/aws/greengrass/mqttbridge/clients/MQTTClient.java
```diff
@@ -117,7 +117,9 @@ protected MQTTClient(Topics topics, MQTTClientKeyStore mqttClientKeyStore, Execu
         this.clientId = Coerce.toString(
                 topics.findOrDefault(DEFAULT_CLIENT_ID, KernelConfigResolver.CONFIGURATION_CONFIG_KEY, CLIENT_ID_KEY));
         this.mqttClientKeyStore = mqttClientKeyStore;
-        this.mqttClientKeyStore.listenToUpdates(this::reset);
+        if (this.serverUri.startsWith("ssl")) {
```
MikeDombo, line 120: Again here don't use starts with.

src/main/java/com/aws/greengrass/mqttbridge/MQTTBridge.java
jcosentino11: Yeah, there's TODO for this in `MQTTClient`, didn't know if it would be better to split out into its own PR, but can add that here 👍

src/main/java/com/aws/greengrass/mqttbridge/MQTTBridge.java
jcosentino11: will take a closer look 👍

PR #56, opened 2022-03-10
refactor: cleanup bridge configuration and handle config changes

jcosentino11:
**Issue #, if available:**

**Description of changes:**

##### Factor out configuration-related constants/code
*  Introduces a `BridgeConfig` class to hold constants and config-getting code

##### Reload on config changes
* listen to configuration topic changes and reinstall as necessary
* To allow safe reload, unsubscribe all listeners during shutdown
    * unsubscribe from CDA's CA topic
    * unsubscribe from TopicMapping changes
* Explicitly start/stop `MessageBridge` within `MQTTBridge` lifecycle

* To allow for cleaner implementation of #58 and #59
* To enable bridge to cleanly restart on configuration change

```
mvn clean verify
```


src/main/java/com/aws/greengrass/mqttbridge/BridgeConfig.java
```diff
@@ -0,0 +1,67 @@
...
+public final class BridgeConfig {
+
+    @Deprecated
+    public static final String KEY_DEPRECATED_BROKER_URI = "brokerServerUri";
+    public static final String KEY_BROKER_URI = "brokerUri";
+    public static final String KEY_CLIENT_ID = "clientId";
+    public static final String KEY_MQTT_TOPIC_MAPPING = "mqttTopicMapping";
+
+    public static final String[] PATH_DEPRECATED_BROKER_URI =
+            {KernelConfigResolver.CONFIGURATION_CONFIG_KEY, KEY_DEPRECATED_BROKER_URI};
```
MikeDombo, line 28: Why do we have a deprecated field? I agree that brokerUri is better, but deprecating a field is nontrivial

src/main/java/com/aws/greengrass/mqttbridge/BridgeConfig.java
```diff
@@ -0,0 +1,67 @@
...
+import java.net.URI;
+import java.net.URISyntaxException;
+
+/**
+ * Utilities for accessing MQTT Bridge configuration.
+ */
+public final class BridgeConfig {
+
+    @Deprecated
+    public static final String KEY_DEPRECATED_BROKER_URI = "brokerServerUri";
```
MikeDombo, line 22: keep constants as private as possible. I haven't read the rest of the change, but if these are only used in this file they should be private.

src/main/java/com/aws/greengrass/mqttbridge/clients/MQTTClient.java
```diff
@@ -92,30 +88,27 @@ public void deliveryComplete(IMqttDeliveryToken token) {
...
-        String tmpUri = Coerce.toString(
-                topics.findOrDefault(DEFAULT_BROKER_URI, KernelConfigResolver.CONFIGURATION_CONFIG_KEY,
-                        DEPRECATED_BROKER_URI_KEY));
-        this.serverUri = Coerce.toString(
-                topics.findOrDefault(tmpUri, KernelConfigResolver.CONFIGURATION_CONFIG_KEY,
-                        BROKER_URI_KEY));
+    protected MQTTClient(Topics topics, MQTTClientKeyStore mqttClientKeyStore,
+                         ExecutorService executorService, IMqttClient mqttClient) throws MQTTClientException {
+        try {
+            this.brokerUri = BridgeConfig.getBrokerUri(topics);
```
MikeDombo, line 105: if the URI changes are you going to recreate this whole MQTTClient instance? That's probably a reasonable solution if that's you plan.

src/main/java/com/aws/greengrass/mqttbridge/clients/MQTTClient.java
```diff
@@ -92,30 +88,27 @@ public void deliveryComplete(IMqttDeliveryToken token) {
...
+        } catch (URISyntaxException e) {
+            throw new MQTTClientException("Invalid brokerUri configuration", e);
```
MikeDombo, line 107: be careful about throwing, you need to ensure that whoever catches the exception will handle it nicely. For example, if the catcher uses `serviceErrored` that will cause the service to go into the errored state and retry. Since it is retrying with the same data we know it will fail which means that it will go into the broken state. Do we handle the broken state correctly? Meaning that if the config changes to a proper URI, do we exit from broken? That isn't guaranteed to work correctly depending on how the rest of this plugin is written.

src/main/java/com/aws/greengrass/mqttbridge/BridgeConfig.java
jcosentino11: according to https://github.com/aws-greengrass/aws-greengrass-mqtt-bridge/pull/48, looks like `brokerServerUri` was once removed but added back for backward compatibility.   

the motivation behind marking it as `@deprecated` is to discourage usage

src/main/java/com/aws/greengrass/mqttbridge/clients/MQTTClient.java
jcosentino11: haven't thought through it completely yet, but yeah, having bridge component recreate this seems appropriate.

src/main/java/com/aws/greengrass/mqttbridge/BridgeConfig.java
jcosentino11: a few of the constants are used in other classes.  the intent behind this class is to be a one-stop shop for these constants, so made them public to encourage their usage in the future.    I feel like inconsistent visibility would be a bit confusing, but happy to make the adjustments if you think otherwise

src/main/java/com/aws/greengrass/mqttbridge/clients/MQTTClient.java
jcosentino11: Will take a closer look, I saw exceptions being thrown in the other constructor so I let my guard down 😅

src/main/java/com/aws/greengrass/mqttbridge/BridgeConfig.java
jcosentino11: update: reduced the visibility of most of the constants in latest commit

src/main/java/com/aws/greengrass/mqttbridge/clients/MQTTClient.java
jcosentino11: update:  MQTTClient is now recreated on configuration change

src/main/java/com/aws/greengrass/mqttbridge/MQTTBridge.java
```diff
@@ -87,35 +86,52 @@ protected MQTTBridge(Topics topics, TopicMapping topicMapping, MessageBridge mes
...
-                this.config.lookupTopics(KernelConfigResolver.CONFIGURATION_CONFIG_KEY, MQTT_TOPIC_MAPPING);
-
-        mappingConfigTopics.subscribe((whatHappened, node) -> {
-            if (mappingConfigTopics.isEmpty()) {
-                logger.debug("Mapping empty");
-                topicMapping.updateMapping(Collections.emptyMap());
+        // handle configuration changes
+        Topics mappingConfigTopics = topics.lookupTopics(BridgeConfig.PATH_MQTT_TOPIC_MAPPING);
+        topics.lookupTopics(KernelConfigResolver.CONFIGURATION_CONFIG_KEY).subscribe((what, child) -> {
+            // initialization
```
MikeDombo, line 93: in the subscribe, ignore the unimportant changes like GenericExternalService does that I sent you. timestamp for example should be ignored

src/main/java/com/aws/greengrass/mqttbridge/MQTTBridge.java
```diff
@@ -87,35 +86,52 @@ protected MQTTBridge(Topics topics, TopicMapping topicMapping, MessageBridge mes
...
-                logger.atInfo().kv("mapping", mapping).log("Updating mapping");
-                topicMapping.updateMapping(mapping);
-            } catch (IllegalArgumentException e) {
-                logger.atError("Invalid topic mapping").kv("TopicMapping", mappingConfigTopics.toString()).log();
-                // Currently, Nucleus spills all exceptions in std err which junit consider failures
-                serviceErrored(String.format("Invalid topic mapping. %s", e.getMessage()));
+            // otherwise, reinstall to completely refresh this plugin
+            if (Arrays.stream(BridgeConfig.ALL_KEYS).anyMatch(child::childOf)) {
+                logger.atInfo().kv("config-key", child.getName())
+                        .log("configuration change detected, reinstalling bridge");
```
MikeDombo, line 108: don't log this, this will actually happen more than once; you'll be notified for every single key that changed, so maybe 4 times or so this will log unnecessarily. We'll see the component reinstall itself from the logs so I think this is superfluous.

src/main/java/com/aws/greengrass/mqttbridge/MQTTBridge.java
```diff
@@ -87,35 +86,52 @@ protected MQTTBridge(Topics topics, TopicMapping topicMapping, MessageBridge mes
...
+                requestReinstall();
             }
         });
     }
 
+    private void onMqttTopicMappingChange(Topics mappingConfigTopics) {
+        if (mappingConfigTopics.isEmpty()) {
+            logger.debug("Mapping empty");
```
MikeDombo, line 116: not a useful log, get rid of it.

src/main/java/com/aws/greengrass/mqttbridge/MQTTBridge.java
```diff
@@ -87,35 +86,52 @@ protected MQTTBridge(Topics topics, TopicMapping topicMapping, MessageBridge mes
...
+
+        try {
+            Map<String, TopicMapping.MappingEntry> mapping = OBJECT_MAPPER
+                    .convertValue(mappingConfigTopics.toPOJO(),
+                            new TypeReference<Map<String, TopicMapping.MappingEntry>>() {
+                            });
+            logger.atInfo().kv("mapping", mapping).log("Updating mapping");
+            topicMapping.updateMapping(mapping);
+        } catch (IllegalArgumentException e) {
+            logger.atError("Invalid topic mapping").kv("TopicMapping", mappingConfigTopics.toString()).log();
```
MikeDombo, line 126: the thing provided to "atError" is the "eventType" which should be a hyphen separated lowercase string. But in most cases we leave it empty. I'd leave it empty here

src/main/java/com/aws/greengrass/mqttbridge/MQTTBridge.java
```diff
@@ -87,35 +86,52 @@ protected MQTTBridge(Topics topics, TopicMapping topicMapping, MessageBridge mes
...
+            // Currently, Nucleus spills all exceptions in std err which junit consider failures
+            serviceErrored(String.format("Invalid topic mapping. %s", e.getMessage()));
```
MikeDombo, line 128: do not both log error and call service errored. Service errored will log itself. 

Do call service errored with the actual exception. This will cause your test to fail, but if it is expected you can ignore it using this pattern: https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/9b1319fb8aaed52441fe34dc0dd36d53e8e20ede/src/integrationtests/java/com/aws/greengrass/integrationtests/deployment/MultiGroupDeploymentTest.java#L405

src/main/java/com/aws/greengrass/mqttbridge/MQTTBridge.java
```diff
@@ -149,8 +162,8 @@ public void startup() {
         }
 
         try {
-            if (mqttClient == null) {
-                mqttClient = new MQTTClient(this.config, mqttClientKeyStore, this.executorService);
+            if (mqttClient == null || mqttClient.isStopped()) {
```
MikeDombo, line 165: why do we need to create it again if it is stopped? Can't we just start it back up?

src/test/java/com/aws/greengrass/mqttbridge/BridgeConfigTest.java
```diff
@@ -0,0 +1,101 @@
...
+
+import java.net.URI;
+import java.net.URISyntaxException;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+
+class BridgeConfigTest {
```
MikeDombo, line 23: use `@ExtendWith(GGExtensions.class)`

src/main/java/com/aws/greengrass/mqttbridge/MQTTBridge.java
jcosentino11: issue is that the changed config is used in the constructor, so we'd need to recreate.  could recreate everytime, but looking through git history there seemed to be an issue with doing that, https://github.com/aws-greengrass/aws-greengrass-mqtt-bridge/commit/24132cca9c28cbc110ee6148549e28d379106979.  maybe with the recent addition of unsubscribing to `MQTTClientKeyStore`(in https://github.com/aws-greengrass/aws-greengrass-mqtt-bridge/pull/56/commits/d38ed00f5c79569ce00b5ed424df968ea0bf7655) that would fix it

jcosentino11: i did try moving `mqttClientInternal` initialization in `MQTTClient` to its `start` method, but that made unit testing cumbersome because it uses a mock for  `mqttClientInternal`

src/main/java/com/aws/greengrass/mqttbridge/MQTTBridge.java
jcosentino11: updated 👍

src/main/java/com/aws/greengrass/mqttbridge/MQTTBridge.java
```diff
@@ -132,26 +149,20 @@ public void startup() {
                 try {
                     List<String> caPemList = (List<String>) newv.toPOJO();
                     if (Utils.isEmpty(caPemList)) {
-                        logger.debug("CA list null or empty");
```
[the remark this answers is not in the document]
jcosentino11: Removed it because similar to https://github.com/aws-greengrass/aws-greengrass-mqtt-bridge/pull/56#discussion_r825118308, this logs doesn't provide much information.  Happy to add it back in if you think otherwise, or maybe it'd be better to log the number of certs if list is not empty.

src/main/java/com/aws/greengrass/mqttbridge/MQTTBridge.java
```diff
@@ -87,35 +86,53 @@ protected MQTTBridge(Topics topics, TopicMapping topicMapping, MessageBridge mes
...
+            }
+
+            // handle mqtt topic mapping changes dynamically
+            if (child.childOf(BridgeConfig.KEY_MQTT_TOPIC_MAPPING)) {
+                onMqttTopicMappingChange(mappingConfigTopics);
+                return;
+            }
+
+            // otherwise, reinstall to completely refresh this plugin
+            if (Arrays.stream(BridgeConfig.ALL_KEYS).anyMatch(child::childOf)) {
```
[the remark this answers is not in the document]
jcosentino11: My concern was that we'd restart bridge if a user added a property we didn't recognize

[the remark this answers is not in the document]
sushantAtAws: I thought about that but I am more concerned about some property we miss to add. User adding an unknown property is a user error IMO.

src/main/java/com/aws/greengrass/mqttbridge/MQTTBridge.java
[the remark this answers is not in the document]
sushantAtAws: Ya. As long as we can make out of this case from the logs.

src/main/java/com/aws/greengrass/mqttbridge/MQTTBridge.java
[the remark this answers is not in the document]
jcosentino11: Yeah that makes sense, will remove this filter 👍

src/main/java/com/aws/greengrass/mqttbridge/BridgeConfig.java
jcosentino11: Ended up renaming constants to `KEY_BROKER_SERVER_URI`/`PATH_BROKER_SERVER_URI` for clarity and left comment about backwards compatibility

src/main/java/com/aws/greengrass/mqttbridge/MQTTBridge.java
```diff
@@ -108,9 +124,7 @@ protected MQTTBridge(Topics topics, TopicMapping topicMapping, MessageBridge mes
             }
 
             // otherwise, reinstall to completely refresh this plugin
-            if (Arrays.stream(BridgeConfig.ALL_KEYS).anyMatch(child::childOf)) {
-                requestReinstall();
-            }
+            requestReinstall();
```
sushantAtAws, line 127: This seems like a nucleus call to reinstall. Can you confirm that we have good logs telling us why this is happening?
To be safe, I would add a clear log signal before this showing the reason for reinstall.

src/main/java/com/aws/greengrass/mqttbridge/MQTTBridge.java
```diff
@@ -175,24 +197,46 @@ public void startup() {
...
         }
 
-        messageBridge.removeMessageClient(TopicMapping.TopicType.IotCore);
         if (ioTCoreClient != null) {
             ioTCoreClient.stop();
         }
     }
+
+    @SuppressWarnings("unchecked")
+    private void subscribeToCertificateAuthoritiesTopic(Consumer<List<String>> onCaChange) throws ServiceLoadException {
```
sushantAtAws, line 224: nit: `onCAChange`

src/main/java/com/aws/greengrass/mqttbridge/MessageBridge.java
```diff
@@ -64,7 +73,6 @@ public MessageBridge(TopicMapping topicMapping) {
      */
     public void addOrReplaceMessageClient(TopicMapping.TopicType clientType, MessageClient messageClient) {
         messageClientMap.put(clientType, messageClient);
-        updateSubscriptionsForClient(clientType, messageClient);
```
sushantAtAws, line 67: These are ordering changes happening. Not against the changes. lgtm. Lets make sure we have the right test coverage to make sure we don't regress anything.

src/main/java/com/aws/greengrass/mqttbridge/MessageBridge.java
```diff
@@ -51,8 +52,16 @@
      */
     public MessageBridge(TopicMapping topicMapping) {
         this.topicMapping = topicMapping;
-        this.topicMapping.listenToUpdates(this::processMappingAndSubscribe);
-        processMappingAndSubscribe();
+    }
+
+    public void start() {
+        topicMapping.listenToUpdates(updateListener);
+        updateListener.onUpdate();
```
sushantAtAws, line 59: I like calling `processMappingAndSubscribe();` directly better. Because then I don't have to go check what `onUpdate` is doing. The code should read like english as much as possible IMO.

src/main/java/com/aws/greengrass/mqttbridge/clients/MQTTClient.java
```diff
@@ -81,34 +80,33 @@ public void deliveryComplete(IMqttDeliveryToken token) {
...
-                         ExecutorService executorService, IMqttClient mqttClient) throws MQTTClientException {
-        try {
-            this.brokerUri = BridgeConfig.getBrokerUri(topics);
-        } catch (URISyntaxException e) {
-            throw new MQTTClientException("Invalid brokerUri configuration", e);
-        }
+    protected MQTTClient(URI brokerUri, String clientId, MQTTClientKeyStore mqttClientKeyStore,
+                         ExecutorService executorService, IMqttClient mqttClient) {
+        Objects.requireNonNull(brokerUri, "Broker URI cannot be null");
+        Objects.requireNonNull(clientId, "Client ID cannot be null");
```
sushantAtAws, line 105: Can it be null though? At this point in code, this has become a developer error. 
We should catch these Invalid input early (in MqttBridge) and surface them as customer errors.

Are these developer asserts at this point?

sushantAtAws: You can use @NonNull annotations to enforce that. It will throw NPE which means a developer error.

src/main/java/com/aws/greengrass/mqttbridge/MQTTBridge.java
jcosentino11: Currently there aren't logs for why, will add those 👍

src/main/java/com/aws/greengrass/mqttbridge/clients/MQTTClient.java
jcosentino11: yep they're developer asserts, will adjust to use nonnull annotation 👍

src/main/java/com/aws/greengrass/mqttbridge/MessageBridge.java
jcosentino11: sounds good, will double check

src/main/java/com/aws/greengrass/mqttbridge/MessageBridge.java
jcosentino11: agreed, will adjust.   unfortunately for `topicMapping.listenToUpdates(updateListener)` I have to still use `updateListener` so that the proper reference is used when calling `TopicMapping#unsubscribe`

src/main/java/com/aws/greengrass/mqttbridge/clients/MQTTClient.java
```diff
@@ -252,26 +243,31 @@ private synchronized void connectAndSubscribe() throws KeyStoreException {
         //TODO: persistent session could be used
         connOpts.setCleanSession(true);
 
-        if (serverUri.startsWith("ssl")) {
+        if ("ssl".equalsIgnoreCase(brokerUri.getScheme())) {
```
jbutler, line 246: Nice

src/main/java/com/aws/greengrass/mqttbridge/MQTTBridge.java
```diff
@@ -86,36 +101,63 @@ protected MQTTBridge(Topics topics, TopicMapping topicMapping, MessageBridge mes
...
-    }
+        this.mqttClientFactory = () -> new MQTTClient(brokerUri, clientId, mqttClientKeyStore, executorService);
 
-    @Override
-    public void install() {
-        mappingConfigTopics =
-                this.config.lookupTopics(KernelConfigResolver.CONFIGURATION_CONFIG_KEY, MQTT_TOPIC_MAPPING);
+        // handle configuration changes
+        Topics mappingConfigTopics = topics.lookupTopics(BridgeConfig.PATH_MQTT_TOPIC_MAPPING);
+        topics.lookupTopics(KernelConfigResolver.CONFIGURATION_CONFIG_KEY).subscribe((what, child) -> {
```
jbutler, line 108: I'm glad you're subscribing to the configuration key now - that avoids/fixes the issue we're seeing in CDA where subscriptions get dropped after a configuration reset. Looks like the same issue was here as well.

Just curious - what's the reason for moving this to the constructor instead of the install method?

src/main/java/com/aws/greengrass/mqttbridge/clients/MQTTClient.java
```diff
@@ -281,6 +277,7 @@ private void reconnectAndResubscribe() {
                 try {
                     Thread.sleep(waitBeforeRetry * 1000);
                 } catch (InterruptedException er) {
+                    Thread.currentThread().interrupt();
                     LOGGER.atError().setCause(er).log("Failed to reconnect");
```
jbutler, line 281: Customers get scared by error logs. I don't think this is actually an error condition - getting interrupted likely means we're shutting down, right? Maybe we can downgrade to info and say we're aborting due to interruption rather than "Failed to reconnect"

src/main/java/com/aws/greengrass/mqttbridge/clients/MQTTClient.java
```diff
@@ -252,26 +243,31 @@ private synchronized void connectAndSubscribe() throws KeyStoreException {
...
-            LOGGER.atInfo().kv("uri", serverUri).kv(CLIENT_ID_KEY, clientId).log("Connected to broker");
+            LOGGER.atInfo().kv("uri", brokerUri).kv(BridgeConfig.KEY_CLIENT_ID, clientId).log("Connected to broker");
         }
     }
 
     private void reconnectAndResubscribe() {
         int waitBeforeRetry = MIN_WAIT_RETRY_IN_SECONDS;
 
         while (!mqttClientInternal.isConnected()) {
+            // prevent connection attempt when interrupted
```
jbutler, line 266: Nice

jbutler (approved):
LGTM

jbutler:
Looks like some test NPEs on Windows which will need to be fixed before merging

src/test/java/com/aws/greengrass/mqttbridge/BridgeConfigTest.java
```diff
@@ -0,0 +1,104 @@
...
+        String clientId = BridgeConfig.getClientId(topics);
+        assertEquals(CLIENT_ID, clientId);
+    }
+
+    @Test
+    void GIVEN_bridgeConfig_WHEN_clientId_missing_THEN_default_clientId_returned() {
+        String clientId = BridgeConfig.getClientId(topics);
+        assertTrue(clientId.startsWith(CLIENT_ID_PREFIX));
+    }
+}
```
MikeDombo, line 104: new line at EOF

src/main/java/com/aws/greengrass/mqttbridge/util/BatchedSubscriber.java
```diff
@@ -0,0 +1,174 @@
...
+
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.BiConsumer;
+import java.util.function.BiPredicate;
+
+/**
+ * NOTE: This class is copied from nucleus, so that we can maintain compatibility with older versions
```
sushantAtAws, line 24: Can this live in Nucleus actually? What is the customization here?

src/test/java/com/aws/greengrass/mqttbridge/MQTTBridgeTest.java
```diff
@@ -177,6 +219,7 @@ void GIVEN_Greengrass_with_mqtt_bridge_WHEN_valid_mqttTopicMapping_updated_THEN_
         kernel.getContext().runOnPublishQueueAndWait(() -> {
         });
         assertThat(topicMapping.getMapping().size(), is(equalTo(3)));
+        assertFalse(bridgeRestarted.await(2, TimeUnit.SECONDS));
```
sushantAtAws, line 222: Why would this test case cause a restart? I may be missing something.

src/main/java/com/aws/greengrass/mqttbridge/util/BatchedSubscriber.java
jcosentino11: I just readjusted the wording to hopefully make it a bit clearer 👍

sushantAtAws (approved):

jcosentino11:
closing, most of this has already been taken care of. will need to follow up w/ separate PRs for unsubscribing from CDA topic and unsubscribing in TopicMapping

PR #58, opened 2022-03-18
feat: allow username/password for local mqtt connection

jcosentino11:
**Issue #, if available:**

**Description of changes:**

Local mqtt connection can now be configured to accept a username and password, provided by bridge configuration. 

```
services:
    aws.greengrass.clientdevices.mqtt.Bridge:
        configuration:
            brokerUri: 'tcp://localhost:8883'
            brokerConnectionOptions:
                username: user
                password: password
```

To allow customers to connect to a local broker with username and password authentication

```
mvn clean verify
```


src/main/java/com/aws/greengrass/mqttbridge/MQTTBridge.java
```diff
@@ -156,6 +159,8 @@ public void install() {
             return;
         }
         this.clientId = BridgeConfig.getClientId(config);
+        this.username = BridgeConfig.getUsername(config);
+        this.password = BridgeConfig.getPassword(config);
```
sushantAtAws, line 163: We can technically disconnect and reconnect if username/password changes. Did we take a design call to reinstall in such cases?

src/main/java/com/aws/greengrass/mqttbridge/clients/MQTTClient.java
```diff
@@ -84,29 +85,41 @@ public void deliveryComplete(IMqttDeliveryToken token) {
      *
      * @param brokerUri          broker uri
      * @param clientId           client id
+     * @param username           optional username
+     * @param password           optional password
```
sushantAtAws, line 89: Seems like a time to make a client config class for input.

src/test/java/com/aws/greengrass/mqttbridge/MQTTBridgeTest.java
```diff
@@ -102,25 +99,45 @@ void cleanup() {
...
 
     @Test
     void GIVEN_Greengrass_with_mqtt_bridge_WHEN_start_kernel_THEN_bridge_starts_successfully() throws Exception {
         startKernelWithConfig("config.yaml");
     }
 
+    @Test
+    void GIVEN_Greengrass_with_mqtt_bridge_WHEN_invalid_credential_config_is_fixed_THEN_bridge_automatically_reinstalls(ExtensionContext context) throws Exception {
+        ignoreExceptionOfType(context, IllegalArgumentException.class);
+        startKernelWithConfig("config_with_invalid_credentials.yaml", State.BROKEN);
```
sushantAtAws, line 128: Is there a way to assert the reason for the BROKEN state?

src/test/java/com/aws/greengrass/mqttbridge/MQTTBridgeTest.java
```diff
@@ -102,25 +99,45 @@ void cleanup() {
...
+
+        CountDownLatch bridgeRunning = new CountDownLatch(1);
+        kernel.getContext().addGlobalStateChangeListener((GreengrassService service, State was, State newState) -> {
+            if (service.getName().equals(MQTTBridge.SERVICE_NAME) && newState.equals(State.RUNNING)) {
```
sushantAtAws, line 132: Lets also assert that the bridge was reinstalled?

src/main/java/com/aws/greengrass/mqttbridge/MQTTBridge.java
jcosentino11: nope there wasn't any design reason for reinstall, and i agree that disconnect/reconnect would be more efficient.  I figured that it would be simplest to reinstall for now and then open a separate PR for disconnect/reconnect?  I started looking into this earlier and I think we'd have to do some refactor work to ensure robustness (e.g. handle exceptions from connectFuture)

src/test/java/com/aws/greengrass/mqttbridge/MQTTBridgeTest.java
jcosentino11: yeah but it would be a bit of a hack to do so....`GreengrassService` doesn't expose this directly, but we could use reflection to look at `GreengrassService#error`.

MikeDombo: Yes you can absolutely assert on the reason. Use a log listener to listen for the error log

MikeDombo: ex: https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/c2e1d321b1a9cc082dcfe902ff8d03dbe55ca810/src/integrationtests/java/com/aws/greengrass/integrationtests/provisioning/ProvisionFromPluginTest.java#L100

jcosentino11: thanks michael, will add that in 👍

src/main/java/com/aws/greengrass/mqttbridge/MQTTBridge.java
sushantAtAws: That is fine by me. Lets open a tracking ticket for that. Reinstall will cause loss of data which is not the best customer experience but I agree with taking this incrementally.

src/main/java/com/aws/greengrass/mqttbridge/clients/MQTTClient.java
```diff
@@ -80,49 +78,52 @@ public void deliveryComplete(IMqttDeliveryToken token) {
...
+        @NonNull
+        String clientId;
+        @NonNull
+        String username;
+        @NonNull
+        String password;
+        MQTTClientKeyStore mqttClientKeyStore;
+        @NonNull
+        ExecutorService executorService;
+        IMqttClient internalMqttClientOverride;
```
sushantAtAws, line 95: This is weird. I like the protected constructor method better.

src/test/java/com/aws/greengrass/mqttbridge/MQTTBridgeTest.java
```diff
@@ -344,4 +337,29 @@ private void updateConfig(String[] config, String value) throws ServiceLoadExcep
...
+    private CountDownLatch trackServiceStateChanges(State... expectedStates) {
+        final Queue<State> states = new LinkedList<>(Arrays.asList(expectedStates));
+        CountDownLatch stateChangesOccurred = new CountDownLatch(1);
+        kernel.getContext().addGlobalStateChangeListener((GreengrassService service, State was, State newState) -> {
+            if (!service.getName().equals(MQTTBridge.SERVICE_NAME)) {
+                return;
+            }
+            synchronized (states) {
+                if (states.isEmpty()) {
+                    return;
```
sushantAtAws, line 350: Not sure when this can occur unless method is called with an empty list. If it does, we should countdown the latch.

jcosentino11: other way it could occur is if another state transition occurs after all the ones we care about have been reached.  say `trackServiceStateChanges(NEW)` and we reinstall a service and let the test run through startup, then this would also trigger `RUNNING` after `NEW`, which would hit this condition

wuetz:
Hi, when will this feature be available?

MikeDombo:
> Hi, when will this feature be available?

We currently have no plans to release this capability.

PR #62, opened 2022-04-07
fix: unsubscribe to CDA CA conf

jcosentino11:
**Issue #, if available:**

**Description of changes:**

Fixes bug where CDA CA subscriptions accumulate for every call to `install`.  Additionally, CDA subscription code has been consolidated to a class, `CertificateAuthorityChangeHandler`, and has been refactored to use `BatchedSubscriber`.

To prevent duplicate subscriptions to CDA topics

unit tests

### DIFF
--- a/src/main/java/com/aws/greengrass/mqtt/bridge/MQTTBridge.java
+++ b/src/main/java/com/aws/greengrass/mqtt/bridge/MQTTBridge.java
@@ -37,6 +37,7 @@ import java.security.cert.CertificateException;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ExecutorService;
 import javax.inject.Inject;
@@ -162,27 +163,19 @@ public class MQTTBridge extends PluginService {
 
         private BatchedSubscriber subscriber;
 
-        @SuppressWarnings({"unchecked", "PMD.UnusedFormalParameter"})
+        @SuppressWarnings("PMD.UnusedFormalParameter")
         private void onCAChange(WhatHappened what, Set<Node> whatChanged) {
-            Topic caTopic;
-            try {
-                caTopic = findCATopic();
-            } catch (ServiceLoadException e) {
-                serviceErrored(e);
-                return;
-            }
-
-            List<String> caCerts = (List<String>) caTopic.toPOJO();
-            if (Utils.isEmpty(caCerts)) {
-                return;
-            }
-
-            logger.atDebug().kv("numCaCerts", caCerts.size()).log("CA update received");
-            try {
-                mqttClientKeyStore.updateCA(caCerts);
-            } catch (IOException | CertificateException | KeyStoreException e) {
-                serviceErrored(e);
-            }
+            maybeFindCATopic()
+                    .map(Coerce::toStringList)
+                    .filter(certs -> !Utils.isEmpty(certs))
+                    .ifPresent(certs -> {
+                        logger.atDebug().kv("numCaCerts", certs.size()).log("CA update received");
+                        try {
+                            mqttClientKeyStore.updateCA(certs);
+                        } catch (IOException | CertificateException | KeyStoreException e) {
+                            serviceErrored(e);
+                        }
+                    });
         }
 
         /**
@@ -197,6 +190,15 @@ public class MQTTBridge extends PluginService {
                 subscriber = new BatchedSubscriber(findCATopic(), this::onCAChange);
             }
             subscriber.subscribe();
+        }
+
+        private Optional<Topic> maybeFindCATopic() {
+            try {
+                return Optional.of(findCATopic());
+            } catch (ServiceLoadException e) {
+                serviceErrored(e);
+                return Optional.empty();
+            }
         }
 
         private Topic findCATopic() throws ServiceLoadException {
--- a/src/main/java/com/aws/greengrass/mqtt/bridge/MQTTBridge.java
+++ b/src/main/java/com/aws/greengrass/mqtt/bridge/MQTTBridge.java
@@ -113,12 +113,7 @@ public class MQTTBridge extends PluginService {
             return;
         }
 
-        try {
-            certificateAuthorityChangeHandler.start();
-        } catch (ServiceLoadException e) {
-            serviceErrored(e);
-            return;
-        }
+        certificateAuthorityChangeHandler.start();
 
         try {
             localMqttClient = localMqttClientFactory.createLocalMqttClient();
@@ -163,50 +158,36 @@ public class MQTTBridge extends PluginService {
 
         private BatchedSubscriber subscriber;
 
-        @SuppressWarnings("PMD.UnusedFormalParameter")
-        private void onCAChange(WhatHappened what, Set<Node> whatChanged) {
-            maybeFindCATopic()
+        private void onCAChange() {
+            findCATopic()
                     .map(Coerce::toStringList)
                     .filter(certs -> !Utils.isEmpty(certs))
-                    .ifPresent(certs -> {
-                        logger.atDebug().kv("numCaCerts", certs.size()).log("CA update received");
-                        try {
-                            mqttClientKeyStore.updateCA(certs);
-                        } catch (IOException | CertificateException | KeyStoreException e) {
-                            serviceErrored(e);
-                        }
-                    });
+                    .ifPresent(this::updateCA);
+        }
+
+        private void updateCA(List<String> certs) {
+            logger.atDebug().kv("numCaCerts", certs.size()).log("CA update received");
+            try {
+                mqttClientKeyStore.updateCA(certs);
+            } catch (IOException | CertificateException | KeyStoreException e) {
+                serviceErrored(e);
+            }
         }
 
         /**
          * Begin listening and responding to CDA CA changes.
          *
          * <p>This operation is idempotent.
-         *
-         * @throws ServiceLoadException if CDA service could not be loaded
          */
-        public void start() throws ServiceLoadException {
+        public void start() {
             if (subscriber == null) {
-                subscriber = new BatchedSubscriber(findCATopic(), this::onCAChange);
+                Topic caTopic = findCATopic().orElse(null);
+                if (caTopic == null) {
+                    return;
+                }
+                subscriber = new BatchedSubscriber(caTopic, (what, whatChanged) -> onCAChange());
             }
             subscriber.subscribe();
-        }
-
-        private Optional<Topic> maybeFindCATopic() {
-            try {
-                return Optional.of(findCATopic());
-            } catch (ServiceLoadException e) {
-                serviceErrored(e);
-                return Optional.empty();
-            }
-        }
-
-        private Topic findCATopic() throws ServiceLoadException {
-            return kernel
-                    .locate(ClientDevicesAuthService.CLIENT_DEVICES_AUTH_SERVICE_NAME).getConfig()
-                    .lookup(RUNTIME_STORE_NAMESPACE_TOPIC,
-                            ClientDevicesAuthService.CERTIFICATES_KEY,
-                            ClientDevicesAuthService.AUTHORITIES_TOPIC);
         }
 
         /**
@@ -215,6 +196,19 @@ public class MQTTBridge extends PluginService {
         public void stop() {
             if (subscriber != null) {
                 subscriber.unsubscribe();
+            }
+        }
+
+        private Optional<Topic> findCATopic() {
+            try {
+                return Optional.of(kernel
+                        .locate(ClientDevicesAuthService.CLIENT_DEVICES_AUTH_SERVICE_NAME).getConfig()
+                        .lookup(RUNTIME_STORE_NAMESPACE_TOPIC,
+                                ClientDevicesAuthService.CERTIFICATES_KEY,
+                                ClientDevicesAuthService.AUTHORITIES_TOPIC));
+            } catch (ServiceLoadException e) {
+                logger.atWarn().cause(e).log("unable to locate service");
+                return Optional.empty();
             }
         }
     }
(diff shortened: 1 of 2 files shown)


src/main/java/com/aws/greengrass/mqttbridge/MQTTBridge.java
```diff
@@ -179,6 +167,61 @@ public void shutdown() {
...
+        @SuppressWarnings({"unchecked", "PMD.UnusedFormalParameter"})
+        private void onCAChange(WhatHappened what, List<Node> whatChanged) {
+            if (whatChanged.isEmpty()) {
+                return;
+            }
+
+            List<String> caCerts = (List<String>) whatChanged.get(whatChanged.size() - 1).toPOJO();
+            if (Utils.isEmpty(caCerts)) {
+                return;
+            }
```
MikeDombo, line 183: this scares me. Perform a node lookup at this point instead of reading from what node list. Then use `Coerce.toStringList`

src/main/java/com/aws/greengrass/mqttbridge/util/BatchedSubscriber.java
```diff
@@ -48,11 +48,11 @@
             what == WhatHappened.timestampUpdated || what == WhatHappened.interiorAdded;
 
     private final AtomicInteger numRequestedChanges = new AtomicInteger();
-    private final Set<Node> whatChanged = new HashSet<>();
+    private final List<Node> whatChanged = new ArrayList<>();
 
     private final Node node;
     private final BiPredicate<WhatHappened, Node> exclusions;
-    private final BiConsumer<WhatHappened, Set<Node>> callback;
+    private final BiConsumer<WhatHappened, List<Node>> callback;
```
MikeDombo, line 55: A set is more appropriate here

src/main/java/com/aws/greengrass/mqttbridge/util/BatchedSubscriber.java
```diff
@@ -48,11 +48,11 @@
             what == WhatHappened.timestampUpdated || what == WhatHappened.interiorAdded;
 
     private final AtomicInteger numRequestedChanges = new AtomicInteger();
-    private final Set<Node> whatChanged = new HashSet<>();
+    private final List<Node> whatChanged = new ArrayList<>();
```
MikeDombo, line 51: If you want to maintain order, use a LinkedHashSet

src/main/java/com/aws/greengrass/mqttbridge/util/BatchedSubscriber.java
jcosentino11: Yeah, i only switched it to have a way to get the "last changed node", but per your suggestion to do a lookup, this isn't necessary anymore, will switch it back to set 👍

MikeDombo: However, I'm more generally worried about storing references to all the nodes, I think that it will encourage bad behavior. I can't really substantiate that, but it is a feeling that it is wrong.

src/main/java/com/aws/greengrass/mqttbridge/util/BatchedSubscriber.java
jcosentino11: 👍  per other comment, going to switch back to set.

src/main/java/com/aws/greengrass/mqttbridge/MQTTBridge.java
jcosentino11: yeah, I think I was trying to be too slick by avoiding a lookup, will switch to a lookup instead 👍

jcosentino11: fixed

src/main/java/com/aws/greengrass/mqttbridge/MQTTBridge.java
```diff
@@ -179,6 +168,65 @@ public void shutdown() {
...
+        private void onCAChange(WhatHappened what, Set<Node> whatChanged) {
+            Topic caTopic;
+            try {
+                caTopic = findCATopic();
+            } catch (ServiceLoadException e) {
+                serviceErrored(e);
+                return;
+            }
+
+            List<String> caCerts = (List<String>) caTopic.toPOJO();
```
MikeDombo, line 185: don't use topojo here and definitely don't blindly cast.

Use Coerce.toStringList

src/main/java/com/aws/greengrass/mqttbridge/MQTTBridge.java
```diff
@@ -179,6 +168,65 @@ public void shutdown() {
...
+         * @throws ServiceLoadException if CDA service could not be loaded
+         */
+        public void start() throws ServiceLoadException {
+            if (subscriber == null) {
+                subscriber = new BatchedSubscriber(findCATopic(), this::onCAChange);
+            }
+            subscriber.subscribe();
+        }
+
+        private Topic findCATopic() throws ServiceLoadException {
```
MikeDombo, line 212: for a nicer API, I'd probably just catch the error in here, and return null, or an Optional<Topic>

src/main/java/com/aws/greengrass/mqttbridge/MQTTBridge.java
jcosentino11: whoops, yeah will adjust

src/main/java/com/aws/greengrass/mqttbridge/MQTTBridge.java
jcosentino11: we'd need a way to propagate the error up though so we can quickly/cleanly exit `startup` though?  that feels a little convoluted with Optional/null.  just pushed up a commit to try it out, I'm not crazy about it, but wanted to get your thoughts if there's a cleaner way

MikeDombo: We're trying to decouple bridge from CDA ultimately, so I don't think that this should really cause the service to error at all. At this time we know that it can never fail because bridge has a hard dependency on CDA. But I'd like to start the scaffolding for decoupling them by not blowing up if the service doesn't exist.

jcosentino11: Ahh okay that makes sense, will remove the error and consider maybe a warn log

jcosentino11: done

src/main/java/com/aws/greengrass/mqttbridge/MQTTBridge.java
```diff
@@ -179,6 +164,65 @@ public void shutdown() {
...
+
+        private Optional<Topic> findCATopic() {
+            try {
+                return Optional.of(kernel
+                        .locate(ClientDevicesAuthService.CLIENT_DEVICES_AUTH_SERVICE_NAME).getConfig()
+                        .lookup(RUNTIME_STORE_NAMESPACE_TOPIC,
+                                ClientDevicesAuthService.CERTIFICATES_KEY,
+                                ClientDevicesAuthService.AUTHORITIES_TOPIC));
+            } catch (ServiceLoadException e) {
+                logger.atWarn().cause(e).log("unable to locate service");
```
MikeDombo, line 220: the log message should make sense to a customer. What is the impact of this warning, and what should they do about it. 

For example:

"Unable to locate `CLIENT_DEVICES_AUTH_SERVICE_NAME`. MQTT Bridge may be unable to connect to the broker. Ensure that `CLIENT_DEVICES_AUTH_SERVICE_NAME` component is deployed"

jcosentino11: done 👍 makes sense, will keep in mind for next time

MikeDombo (approved):

src/main/java/com/aws/greengrass/mqttbridge/MQTTBridge.java
```diff
@@ -217,7 +217,12 @@ private Optional<Topic> findCATopic() {
                                 ClientDevicesAuthService.CERTIFICATES_KEY,
                                 ClientDevicesAuthService.AUTHORITIES_TOPIC));
             } catch (ServiceLoadException e) {
-                logger.atWarn().cause(e).log("unable to locate service");
+                logger.atWarn().cause(e).log(String.format(
```
MikeDombo, line 220: don't use string format. Our logger has builtin formatting using `{}`. `.log("a {}, b {}", a, b)`

MikeDombo (changes requested):
don't use string format in logging.

src/main/java/com/aws/greengrass/mqttbridge/MQTTBridge.java
jcosentino11: fixed

MikeDombo (approved):

src/main/java/com/aws/greengrass/mqtt/bridge/MQTTBridge.java
```diff
@@ -167,6 +151,70 @@ public void shutdown() {
...
+                mqttClientKeyStore.updateCA(certs);
+            } catch (IOException | CertificateException | KeyStoreException e) {
+                serviceErrored(e);
+            }
+        }
+
+        private Optional<Topic> findCATopic() {
+            try {
+                return Optional.of(kernel
+                        .locate(ClientDevicesAuthService.CLIENT_DEVICES_AUTH_SERVICE_NAME).getConfig()
```
MikeDombo, line 202: just use `getRuntimeConfig`

src/test/java/com/aws/greengrass/mqtt/bridge/MQTTBridgeTest.java
```diff
@@ -332,62 +334,86 @@ void GIVEN_Greengrass_with_mqtt_bridge_WHEN_invalid_mqttTopicMapping_updated_THE
...
-        initializeMockedConfig();
-        TopicMapping mockTopicMapping = mock(TopicMapping.class);
-        MessageBridge mockMessageBridge = mock(MessageBridge.class);
-        PubSubIPCEventStreamAgent mockPubSubIPCAgent = mock(PubSubIPCEventStreamAgent.class);
-        Kernel mockKernel = mock(Kernel.class);
-        MQTTClientKeyStore mockMqttClientKeyStore = mock(MQTTClientKeyStore.class);
-        MQTTBridge mqttBridge;
-        LocalMqttClientFactory localMqttClientFactory = new LocalMqttClientFactory(mockMqttClientKeyStore, ses);
+    void GIVEN_mqtt_bridge_WHEN_cda_ca_conf_changed_THEN_bridge_keystore_updated() throws Exception {
+        startKernelWithConfig("config.yaml");
```
MikeDombo, line 342: starting a kernel makes this an integration test. Try and keep this as a unit test.

src/test/java/com/aws/greengrass/mqtt/bridge/MQTTBridgeTest.java
```diff
@@ -332,62 +334,86 @@ void GIVEN_Greengrass_with_mqtt_bridge_WHEN_invalid_mqttTopicMapping_updated_THE
...
+    void GIVEN_mqtt_bridge_WHEN_cda_ca_conf_changed_after_shutdown_THEN_bridge_keystore_not_updated(ExtensionContext context) throws Exception {
+        ignoreExceptionOfType(context, IllegalArgumentException.class);
+        ignoreExceptionOfType(context, NullPointerException.class);
 
-        try (MqttClient mockIotMqttClient = mock(MqttClient.class)) {
-            mqttBridge =
-                    new MQTTBridge(config, mockTopicMapping, mockMessageBridge, mockPubSubIPCAgent, mockIotMqttClient,
-                            mockKernel, mockMqttClientKeyStore, localMqttClientFactory, ses);
-        }
+        startKernelWithConfig("config.yaml");
```
MikeDombo, line 374: same here

MikeDombo:
don't just retry the windows failure, make sure there's nothing we're missing to fix the flakiness.

src/test/java/com/aws/greengrass/mqtt/bridge/MQTTBridgeTest.java
jcosentino11: yeah, since the tests involve install/startup/shutdown lifecycle, i figured int test would give better assurances than mocking.  can revert it back to unit test though

MikeDombo: yes please. Having an integ test in addition is fine, but it should go with the other integ tests.

src/main/java/com/aws/greengrass/mqtt/bridge/MQTTBridge.java
```diff
@@ -167,6 +151,70 @@ public void shutdown() {
...
+                                ClientDevicesAuthService.CERTIFICATES_KEY,
+                                ClientDevicesAuthService.AUTHORITIES_TOPIC));
+            } catch (ServiceLoadException e) {
+                logger.atWarn().cause(e).log(
+                        "Unable to locate {}. "
+                                + "MQTT Bridge may be unable to connect to the broker. "
+                                + "Ensure that {} component is deployed.",
+                        ClientDevicesAuthService.CLIENT_DEVICES_AUTH_SERVICE_NAME,
+                        ClientDevicesAuthService.CLIENT_DEVICES_AUTH_SERVICE_NAME);
+                return Optional.empty();
```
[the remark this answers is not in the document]
jcosentino11: https://github.com/aws-greengrass/aws-greengrass-mqtt-bridge/pull/62#discussion_r845520851 🤷 fine either way

jcosentino11:
> don't just retry the windows failure, make sure there's nothing we're missing to fix the flakiness.

yep, it was complaining about a client id (that we set in a test different than the one that failed). probably appeared because we have a mix of int test and mocked `GGServiceTestUtil` stuff in same class

src/main/java/com/aws/greengrass/mqtt/bridge/MQTTBridge.java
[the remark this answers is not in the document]
MikeDombo: ok then, thanks for reminding me, that's why my comment wasn't saying it is wrong :)
Just looking for an explanation.

src/test/java/com/aws/greengrass/mqtt/bridge/MQTTBridgeTest.java
jcosentino11: yeah, we need a refactor, cause there's no integ folder for bridge yet.  they're just in this `MQTTBridgeTest` class

MikeDombo (approved):

robcmann (approved):

PR #64, opened 2022-04-08
fix: remove client certificate rotate callback

jbutler:
It is not necessary to disconnect the MQTT Bridge when the client
certificate rotates. This change removes that logic, but still ensures
that the new client certificate is used the next time the bridge
connects.

CA rotation still results in a disconnect, as that indicates that the
client should no longer trust the server.

**Issue #, if available:**

**Description of changes:**

### DIFF
--- a/src/main/java/com/aws/greengrass/mqttbridge/auth/MQTTClientKeyStore.java
+++ b/src/main/java/com/aws/greengrass/mqttbridge/auth/MQTTClientKeyStore.java
@@ -56,7 +56,7 @@ public class MQTTClientKeyStore {
 
     @FunctionalInterface
     public interface UpdateListener {
-        void onUpdate();
+        void onCAUpdate();
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/main/java/com/aws/greengrass/mqttbridge/auth/MQTTClientKeyStore.java
```diff
@@ -109,8 +109,6 @@ private KeyPair newRSAKeyPair() throws NoSuchAlgorithmException {
     private void updateCert(X509Certificate... certChain) {
         try {
             keyStore.setKeyEntry(KEY_ALIAS, keyPair.getPrivate(), DEFAULT_KEYSTORE_PASSWORD, certChain);
-
-            updateListeners.forEach(UpdateListener::onUpdate); //notify MQTTClient
```
sushantAtAws, line 113: Can we atleast add a log here?

sushantAtAws (approved):

src/main/java/com/aws/greengrass/mqttbridge/auth/MQTTClientKeyStore.java
jbutler: Done

vaibhavmurkute (approved):

sushantAtAws (approved):

jcosentino11 (approved):

Issue #68, opened 2022-06-16
MQTT Bridge: encapsulate bridged traffic when transporting through IPC


cBiscuitSurprise:
**Feature Description**
Our company needs separation of concerns when it comes to transporting messages over IPC. MQTT-topics and IPC-topics serve two different purposes.

**Use Case**
Devices communicating over arbitrary manufacturer-specific MQTT topics need to be routed to a common Greengrass Component via IPC, where the IPC input/output topics are independent of the MQTT topics.

**Proposed Solution**
The MQTT payload (including it's topic) could be encapsulated within the IPC payload and transmitted over a separately-configured IPC topic. This way, the original MQTT context is preserved, but the IPC transport can be managed separately.

**Current MQTT Bridge Configuration**
```json
"mqttTopicMapping": {
  "ClientDevice1Mapping": {
    "topic": "clients/MyClientDevice1/hello/world",
    "source": "LocalMqtt",
    "target": "IotCore"
  },
  "ClientDevice2Mapping": {
    "topic": "clients/MyClientDevice2/hello/world",
    "source": "LocalMqtt",
    "target": "IotCore"
  }
}
```

**Conceptual proposed MQTT Bridge Configuration**
```json
"mqttTopicMapping": {
  "ClientDevice1Mapping": {
    "sourceTopic": "clients/MyClientDevice1/hello/world",
    "source": "LocalMqtt",
    "targetTopic": "connector/mqtts/message/input",
    "target": "Pubsub"
  },
  "ClientDevice2Mapping": {
    "sourceTopic": "clients/MyClientDevice2/hello/world",
    "source": "LocalMqtt",
    "targetTopic": "connector/mqtts/message/input",
    "target": "Pubsub"
  }
  "OutboundMapping": {
    "sourceTopic": "connector/mqtts/message/output",
    "source": "Pubsub",
    "targetTopic": "clients/MyClientDevice2/command",
    "target": "LocalMqtt"
  }
}
```

But then, rather than just `message.getPayload()` being transported, the IPC payload would be something like:
```json
{
  "mqttTopic": "clients/MyClientDevice2/hello/world",
  "mqttPayload": message.getPayload()
}
```

This would only apply to source/target = `Pubsub`, if `Pubsub` isn't involved, then proceed as implemented today.

**Other**

* [x] :wave: I may be able to implement this feature request
* [x] :warning: This feature might incur a breaking change
[...]

PR #69, opened 2022-06-16
Enable configuration of `targetTopic` independently of `sourceTopic`

cBiscuitSurprise:
**Issue #, if available:**
GH-68

**Description of changes:**
This PR just implements the ability to configure target-topics independently from source-topics. The payload is bridged un-changed (I intend to submit a future PR adding the ability to define the target-payload).

This change helps to separate the configuration different transports during the bridge process (organization of IPC topics is generally not governed by the same rules as that of MQTT topics).

Unit tests only (I haven't built/deployed this yet).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+greengrass-build/
 **/target/
 .DS_Store
 ._*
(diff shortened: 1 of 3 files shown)


cBiscuitSurprise:
For reference, I moved this work to [`cBiscuitSurprise:addTargetTopic`](https://github.com/cBiscuitSurprise/aws-greengrass-mqtt-bridge/tree/addTargetTopic), but can't update this PR to reflect that update.

PR #71, opened 2022-06-20
chore: migrate to subscribeToCertificateUpdates API

jbutler:
**Issue #, if available:**

**Description of changes:**
Removes CSR generation logic for getting client certificates and instead uses the subscribeToCertificateUpdates API

Depends on https://github.com/aws-greengrass/aws-greengrass-client-device-auth/pull/95
CI will fail until the above is merged.

### DIFF
--- a/src/main/java/com/aws/greengrass/mqttbridge/auth/MQTTClientKeyStore.java
+++ b/src/main/java/com/aws/greengrass/mqttbridge/auth/MQTTClientKeyStore.java
@@ -5,22 +5,22 @@
 
 package com.aws.greengrass.mqttbridge.auth;
 
-import com.aws.greengrass.certificatemanager.CertificateManager;
-import com.aws.greengrass.certificatemanager.certificate.CertificateRequestGenerator;
-import com.aws.greengrass.certificatemanager.certificate.CsrProcessingException;
+import com.aws.greengrass.device.ClientDevicesAuthServiceApi;
+import com.aws.greengrass.device.api.CertificateUpdateEvent;
+import com.aws.greengrass.device.api.GetCertificateRequest;
+import com.aws.greengrass.device.api.GetCertificateRequestOptions;
+import com.aws.greengrass.device.exception.CertificateGenerationException;
 import com.aws.greengrass.logging.api.Logger;
 import com.aws.greengrass.logging.impl.LogManager;
+import com.aws.greengrass.mqttbridge.MQTTBridge;
 import lombok.AccessLevel;
 import lombok.Getter;
-import org.bouncycastle.operator.OperatorCreationException;
 
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
 import java.security.KeyManagementException;
-import java.security.KeyPair;
-import java.security.KeyPairGenerator;
 import java.security.KeyStore;
 import java.security.KeyStoreException;
 import java.security.NoSuchAlgorithmException;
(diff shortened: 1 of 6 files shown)


src/main/java/com/aws/greengrass/mqttbridge/auth/MQTTClientKeyStore.java
```diff
@@ -62,56 +59,45 @@ public interface UpdateListener {
...
-            keyStore.setKeyEntry(KEY_ALIAS, keyPair.getPrivate(), DEFAULT_KEYSTORE_PASSWORD, certChain);
+            X509Certificate[] certChain = Stream.concat(
+                            Stream.of(certificateUpdate.getCertificate()),
+                            Arrays.stream(certificateUpdate.getCaCertificates()))
+                    .toArray(X509Certificate[]::new);
+            keyStore.setKeyEntry(
+                    KEY_ALIAS, certificateUpdate.getKeyPair().getPrivate(), DEFAULT_KEYSTORE_PASSWORD, certChain);
         } catch (KeyStoreException e) {
-            LOGGER.atError("Unable to store generated cert", e);
+            LOGGER.atError().log("Unable to store generated cert", e);
```
MikeDombo, line 100: wow, good fix

src/main/java/com/aws/greengrass/mqttbridge/auth/MQTTClientKeyStore.java
```diff
@@ -62,56 +59,45 @@ public interface UpdateListener {
...
-
-    private KeyPair newRSAKeyPair() throws NoSuchAlgorithmException {
-        KeyPairGenerator kpg = KeyPairGenerator.getInstance(RSA_KEY_INSTANCE);
-        kpg.initialize(RSA_KEY_LENGTH);
-        return kpg.generateKeyPair();
+        GetCertificateRequestOptions options = new GetCertificateRequestOptions();
+        options.setCertificateType(GetCertificateRequestOptions.CertificateType.CLIENT);
+        GetCertificateRequest certificateRequest =
+                new GetCertificateRequest(MQTTBridge.SERVICE_NAME, options, this::updateCert);
+        clientDevicesAuthServiceApi.subscribeToCertificateUpdates(certificateRequest);
```
MikeDombo, line 87: unsubscribe on shutdown if possible

MikeDombo (dismissed):

src/main/java/com/aws/greengrass/mqttbridge/auth/MQTTClientKeyStore.java
jbutler: Good catch, this needs to be addressed

src/main/java/com/aws/greengrass/mqttbridge/auth/MQTTClientKeyStore.java
```diff
@@ -80,11 +82,14 @@ public void init() throws KeyStoreException, CertificateGenerationException {
             throw new KeyStoreException("Unable to load keystore", e);
         }
 
-        GetCertificateRequestOptions options = new GetCertificateRequestOptions();
-        options.setCertificateType(GetCertificateRequestOptions.CertificateType.CLIENT);
-        GetCertificateRequest certificateRequest =
-                new GetCertificateRequest(MQTTBridge.SERVICE_NAME, options, this::updateCert);
-        clientDevicesAuthServiceApi.subscribeToCertificateUpdates(certificateRequest);
+        clientDevicesAuthServiceApi.subscribeToCertificateUpdates(clientCertificateRequest);
```
MikeDombo, line 85: this could be OK, but it is a bit asymmetric. Since you're subscribing from the constructor, but unsubscribing in shutdown. What if it was restarted, would it break?

MikeDombo: ok, got it with the diff showing more context. That's fine then.

MikeDombo (approved):

jcosentino11 (approved):

PR #75, opened 2022-07-19
fix: retry IotCoreClient subscriptions when greengrass starts offline

jcosentino11:
**Issue #, if available:**

**Description of changes:**

* Adjust the behavior of `IotCoreClient` so that subscriptions are retried in the event greengrass was started offline.
* Return early if interrupted in `IotCoreClient#subscribeToTopicsWithRetry`

Customers may start greengrass while offline so we must handle that scenario in `IotCoreClient`

Unit test and UAT

### DIFF
--- a/src/main/java/com/aws/greengrass/mqtt/bridge/clients/IoTCoreClient.java
+++ b/src/main/java/com/aws/greengrass/mqtt/bridge/clients/IoTCoreClient.java
@@ -49,7 +49,6 @@ public class IoTCoreClient implements MessageClient {
     @Getter(AccessLevel.PACKAGE) // for unit testing
     private Future<?> subscribeFuture;
     private final Object subscribeLock = new Object();
-    private final AtomicBoolean connectionInterrupted = new AtomicBoolean();
     private final MqttClient iotMqttClient;
     private final ExecutorService executorService;
 
@@ -70,12 +69,10 @@ public class IoTCoreClient implements MessageClient {
     private final MqttClientConnectionEvents connectionCallbacks = new MqttClientConnectionEvents() {
         @Override
         public void onConnectionInterrupted(int errorCode) {
-            connectionInterrupted.set(true);
         }
 
         @Override
         public void onConnectionResumed(boolean sessionPresent) {
-            connectionInterrupted.set(false);
             synchronized (subscribeLock)  {
                 if (subscribeFuture != null) {
                     subscribeFuture.cancel(true);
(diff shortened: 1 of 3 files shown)


src/main/java/com/aws/greengrass/mqtt/bridge/clients/IoTCoreClient.java
```diff
@@ -29,6 +29,7 @@
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicBoolean;
```
MikeDombo, line 32: remove unused

jcosentino11: yep just fixed

src/test/java/com/aws/greengrass/mqtt/bridge/clients/IoTCoreClientTest.java
```diff
@@ -79,19 +77,31 @@ void GIVEN_iotcore_client_started_WHEN_update_subscriptions_THEN_topics_subscrib
...
         Set<String> topics = new HashSet<>();
         topics.add("iotcore/topic");
         topics.add("iotcore/topic2");
+
+        // attempt to update subscriptions, this is expected to fail since bridge is offline
+        when(mockIotMqttClient.connected()).thenReturn(false);
         iotCoreClient.updateSubscriptions(topics, message -> {
         });
 
-        verify(mockIotMqttClient, never()).subscribe(any(SubscribeRequest.class));
```
MikeDombo, line 92: why removed?

MikeDombo (approved):
Since this has exposed an unknown mode of operation, we need a test which verifies the retry functionality when offline from the beginning (which the UAT does) as well as when it is connected and then goes offline, then back online.

src/main/java/com/aws/greengrass/mqtt/bridge/clients/IoTCoreClient.java
```diff
@@ -44,26 +44,26 @@ public class IoTCoreClient implements MessageClient {
     private Set<String> subscribedIotCoreTopics = ConcurrentHashMap.newKeySet();
     @Getter(AccessLevel.PROTECTED)
     private Set<String> toSubscribeIotCoreTopics = new HashSet<>();
-
-    private Consumer<Message> messageHandler;
+    private volatile Consumer<Message> messageHandler;
```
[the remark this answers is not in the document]
jcosentino11: it's set every time `updateSubscriptions` is called, but yeah it doesn't matter all that much right now since the lambda content doesn't change `message -> handleMessage(clientType, message)`, but it's technically a new reference each time

vaibhavmurkute (approved):

PR #76, opened 2022-08-05
fix(logging): drop message sent/received to debug level

MikeDombo:
**Issue #, if available:**

**Description of changes:**
Change the log level in the message bridge for each received and sent message down to debug. It was logging at info for every single message which is forwarded which can have an enormous volume.

### DIFF
--- a/src/main/java/com/aws/greengrass/mqtt/bridge/MessageBridge.java
+++ b/src/main/java/com/aws/greengrass/mqtt/bridge/MessageBridge.java
@@ -111,7 +111,7 @@ public class MessageBridge {
                     Message msg = new Message(targetTopic, message.getPayload());
                     try {
                         client.publish(msg);
-                        LOGGER.atInfo().kv(LOG_KEY_SOURCE_TYPE, sourceType).kv(LOG_KEY_SOURCE_TOPIC, fullSourceTopic)
+                        LOGGER.atDebug().kv(LOG_KEY_SOURCE_TYPE, sourceType).kv(LOG_KEY_SOURCE_TOPIC, fullSourceTopic)
                                 .kv(LOG_KEY_TARGET_TYPE, mapping.getTarget())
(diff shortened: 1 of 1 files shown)


src/main/java/com/aws/greengrass/mqtt/bridge/MessageBridge.java
```diff
@@ -111,7 +111,7 @@ private void handleMessage(TopicMapping.TopicType sourceType, Message message) {
                     Message msg = new Message(targetTopic, message.getPayload());
                     try {
                         client.publish(msg);
-                        LOGGER.atInfo().kv(LOG_KEY_SOURCE_TYPE, sourceType).kv(LOG_KEY_SOURCE_TOPIC, fullSourceTopic)
+                        LOGGER.atTrace().kv(LOG_KEY_SOURCE_TYPE, sourceType).kv(LOG_KEY_SOURCE_TOPIC, fullSourceTopic)
```
jbutler, line 114: Trace is too low. Debug is fine for this

jcosentino11 (approved):

src/main/java/com/aws/greengrass/mqtt/bridge/MessageBridge.java
MikeDombo: I wouldn't really agree here. Trace should be for tracing individual messages (which is what this log is). Debug can be more commonly enabled and it should provide more details, but not necessarily more frequency.

vaibhavmurkute (dismissed):

src/main/java/com/aws/greengrass/mqtt/bridge/MessageBridge.java
jbutler: We need visibility into this when customers aren't seeing their messages flowing how they expect. Trace isn't even a documented log level.

Needs to be debug

jcosentino11 (approved):

jbutler (approved):

PR #81, opened 2023-03-09
feat: mqtt5 config flag

jcosentino11:
**Issue #, if available:**

**Description of changes:**

Adds a `configuration.brokerClient.version` config.  Possible values are `mqtt3` and `mqtt5` with a default of `mqtt3`.  This config will control which local mqtt client is used (v3 or v5).  If this property changes, bridge will be reinstalled.

To support configurable mqtt3/mqtt5 within bridge

unit tests

### DIFF
--- a/src/main/java/com/aws/greengrass/mqtt/bridge/BridgeConfig.java
+++ b/src/main/java/com/aws/greengrass/mqtt/bridge/BridgeConfig.java
@@ -45,14 +45,6 @@ public final class BridgeConfig {
     private final String clientId;
     private final Map<String, TopicMapping.MappingEntry> topicMapping;
 
-    /**
-     * Create a BridgeConfig from configuration topics.
-     *
-     * @param configurationTopics configuration topics
-     * @return bridge config
-     * @throws InvalidConfigurationException if config from topics is invalid
-     */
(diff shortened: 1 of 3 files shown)


src/main/java/com/aws/greengrass/mqtt/bridge/BridgeConfig.java
```diff
@@ -54,44 +59,64 @@ public final class BridgeConfig {
...
+            throw new InvalidConfigurationException("Malformed " + KEY_MQTT_TOPIC_MAPPING, e);
+        }
     }
 
+    private static MqttVersion getMqttVersion(Topics configurationTopics) throws InvalidConfigurationException {
+        String versionName = Coerce.toString(configurationTopics.findOrDefault(
+                DEFAULT_MQTT_VERSION,
+                KEY_BROKER_CLIENT, KEY_VERSION));
+        MqttVersion version = MqttVersion.fromName(versionName);
+        if (version == null) {
```
MikeDombo, line 104: if the version is null, use the default.

sameerzuberi (approved):

robcmann (approved):

PR #82, opened 2023-03-09
feat: local mqtt client factory

jcosentino11:
**Issue #, if available:**

**Description of changes:**

Adds a `LocalMqttClientFactory`, to allow bridge to choose between mqtt3 and 5 clients, based on configuration.

To support v3 and v5 local mqtt clients

unit tests

### DIFF
--- a/src/main/java/com/aws/greengrass/mqtt/bridge/MQTTBridge.java
+++ b/src/main/java/com/aws/greengrass/mqtt/bridge/MQTTBridge.java
@@ -17,8 +17,9 @@ import com.aws.greengrass.lifecyclemanager.PluginService;
 import com.aws.greengrass.lifecyclemanager.exceptions.ServiceLoadException;
 import com.aws.greengrass.mqtt.bridge.auth.MQTTClientKeyStore;
 import com.aws.greengrass.mqtt.bridge.clients.IoTCoreClient;
-import com.aws.greengrass.mqtt.bridge.clients.MQTTClient;
+import com.aws.greengrass.mqtt.bridge.clients.LocalMqttClientFactory;
 import com.aws.greengrass.mqtt.bridge.clients.MQTTClientException;
+import com.aws.greengrass.mqtt.bridge.clients.MessageClient;
 import com.aws.greengrass.mqtt.bridge.clients.PubSubClient;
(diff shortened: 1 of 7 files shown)


src/main/java/com/aws/greengrass/mqtt/bridge/MQTTBridge.java
```diff
@@ -125,10 +130,10 @@ public void startup() {
         }
 
         try {
-            mqttClient = new MQTTClient(bridgeConfig.getBrokerUri(), bridgeConfig.getClientId(),
-                    mqttClientKeyStore, executorService);
-            mqttClient.start();
-            messageBridge.addOrReplaceMessageClientAndUpdateSubscriptions(TopicMapping.TopicType.LocalMqtt, mqttClient);
+            localMqttClient = localMqttClientFactory.createLocalMqttClient();
```
MikeDombo, line 133: this will not currently enable switching versions right? This code would only be called once right now I believe.

If so, that's fine, just want to make sure I'm on the same page.

jcosentino11: when version conf is changed, bridge will reinstall and this would get called again

jcosentino11: https://github.com/aws-greengrass/aws-greengrass-mqtt-bridge/blob/1540f3c7c0dfc6801da97a5eaa8259bde216a31e/src/main/java/com/aws/greengrass/mqtt/bridge/BridgeConfig.java#L119

MikeDombo: OK, that's not ideal because restart/reinstall will stop the bridge which will have it unsubscribe from everything including cloud topics which we have no reason to unsubscribe from just to change our mqtt version.

I'm ok with this for now, but consider how we may do it dynamically.

jcosentino11: yeah that's fair, will do as a follow-up 👍

sameerzuberi (approved):

PR #84, opened 2023-03-09
chore: unsubscribe from keystore subscription

jcosentino11:
**Issue #, if available:**

**Description of changes:**
Unsubscribe from keystore operations within MQTTClient

Cleaner shutdown

unit tests

### DIFF
--- a/src/main/java/com/aws/greengrass/mqtt/bridge/auth/MQTTClientKeyStore.java
+++ b/src/main/java/com/aws/greengrass/mqtt/bridge/auth/MQTTClientKeyStore.java
@@ -151,6 +151,14 @@ public class MQTTClientKeyStore {
     }
 
     /**
+     * Stop listening to KeyStore updates for a given listener.
+     * @param listener callback
+     */
+    public synchronized void unsubscribeToUpdates(UpdateListener listener) {
+        updateListeners.remove(listener);
+    }
+
+    /**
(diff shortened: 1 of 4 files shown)


sameerzuberi (approved):

MikeDombo (approved):

src/main/java/com/aws/greengrass/mqtt/bridge/auth/MQTTClientKeyStore.java
```diff
@@ -150,6 +150,14 @@ public synchronized void listenToCAUpdates(UpdateListener listener) {
         updateListeners.add(listener);
     }
 
+    /**
+     * Stop listening to KeyStore updates for a given listener.
+     * @param listener callback
+     */
+    public synchronized void unsubscribeToUpdates(UpdateListener listener) {
```
MikeDombo, line 157: Same as here: https://github.com/aws-greengrass/aws-greengrass-mqtt-bridge/pull/85/files#r1132380603.

You made a good change the the MQTTClient file so that the subscribe won't change, and I think that's sufficient. I don't think we need or actually want to unsubscribe.

robcmann (approved):

jcosentino11:
went with https://github.com/aws-greengrass/aws-greengrass-mqtt-bridge/pull/115 instead

PR #85, opened 2023-03-09
chore: unsubscribe from TopicMapping updates

jcosentino11:
**Issue #, if available:**

**Description of changes:**
Have message bridge unsubscribe from topic mapping changes during shutdown

Ensure clean shutdown

unit tests

### DIFF
--- a/src/main/java/com/aws/greengrass/mqtt/bridge/MessageBridge.java
+++ b/src/main/java/com/aws/greengrass/mqtt/bridge/MessageBridge.java
@@ -70,6 +70,7 @@ public class MessageBridge {
      */
     public void stop() {
         topicMapping.unsubscribeFromUpdates(updateListener);
+        messageClientMap.clear();
     }
 
     /**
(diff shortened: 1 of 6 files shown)


sameerzuberi (approved):

src/main/java/com/aws/greengrass/mqtt/bridge/MessageBridge.java
```diff
@@ -50,21 +51,38 @@ public class MessageBridge {
...
+    public void start() {
+        topicMapping.listenToUpdates(updateListener);
         processMappingAndSubscribe();
     }
 
+    /**
+     * Stop the message bridge.
+     */
+    public void stop() {
+        topicMapping.unsubscribeFromUpdates(updateListener);
```
MikeDombo, line 72: not clear to me that we do want to unsubscribe here. I feel that there's a potential for missing changes if we do this. As long as you only subscribe once or even if you subscribe multiple times but with the same listener, there won't be multiple subscriptions.

robcmann (approved):

PR #88, opened 2023-03-10
refactor: convert Message to interface

jcosentino11:
**Issue #, if available:**

**Description of changes:**
Convert `Message` from a class to an interface, with the ability for it to convert itself to mqtt or pubsub types.

To keep `Message` generic without cluttering it with Mqtt5 fields, which will be added in the near future

unit tests


src/main/java/com/aws/greengrass/mqtt/bridge/model/Message.java
```diff
@@ -0,0 +1,21 @@
...
+public interface Message {
+    Message toPubSub();
+
+    Message toMqtt();
+
+    String getTopic();
+
+    byte[] getPayload();
+
+    Message ofTopic(String topic);
```
MikeDombo, line 20: why does this exist? We already have a `Message`, so why are we then calling a method on it to give us a different message which a presumably different topic?

src/main/java/com/aws/greengrass/mqtt/bridge/model/MqttMessage.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+package com.aws.greengrass.mqtt.bridge.model;
+
+import lombok.NonNull;
+import lombok.Value;
+
+// TODO support mqtt5
+@Value
+public class MqttMessage implements Message {
```
MikeDombo, line 13: Rename to LocalMqttMessage

src/main/java/com/aws/greengrass/mqtt/bridge/model/Message.java
jcosentino11: Yeah, cause of generics weirdness https://github.com/aws-greengrass/aws-greengrass-mqtt-bridge/pull/88/files#diff-7d623726e13974aaebfada082983f1677688fcc2f9225dc9a51c63a580ee89c9R145-R150.  I guess could just make it a setTopic, but that would remove the immutability of message

src/main/java/com/aws/greengrass/mqtt/bridge/model/MqttMessage.java
jcosentino11: Was also using this in iot core client, did we want a separate type for that?

src/main/java/com/aws/greengrass/mqtt/bridge/model/Message.java
MikeDombo: I think the name is confusing. How about `newFromMessageWithTopic`?

jcosentino11: yeah will change

jcosentino11: done

MikeDombo (approved):

robcmann (approved):

PR #98, opened 2023-03-17
refactor: cleanup exception handling

jcosentino11:
**Issue #, if available:**

**Description of changes:**

Refactor so that bridge catches more generic `MessageClientException` instead of `MQTTClientException`, and methods throw `MessageClientException` rather than `MQTTClientException`.

Cleanups unnecessary `throws` statements in  `MQTTClient` 

### DIFF
--- a/src/main/java/com/aws/greengrass/mqtt/bridge/clients/MQTTClient.java
+++ b/src/main/java/com/aws/greengrass/mqtt/bridge/clients/MQTTClient.java
@@ -100,7 +100,6 @@ public class MQTTClient implements MessageClient<MqttMessage> {
     public MQTTClient(@NonNull URI brokerUri, @NonNull String clientId, MQTTClientKeyStore mqttClientKeyStore,
                       ExecutorService executorService) throws MessageClientException {
         this(brokerUri, clientId, mqttClientKeyStore, executorService, null);
-        // TODO: Handle the case when serverUri is modified
         try {
             this.mqttClientInternal = new MqttClient(brokerUri.toString(), clientId, dataStore);
         } catch (MqttException e) {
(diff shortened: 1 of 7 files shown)


src/main/java/com/aws/greengrass/mqtt/bridge/clients/MQTTClient.java
```diff
@@ -142,12 +142,12 @@ void reset() {
      * @throws RuntimeException if the client cannot load the KeyStore used to connect to the broker.
      */
     @Override
```
MikeDombo, line 144: fix the throws documentation

src/main/java/com/aws/greengrass/mqtt/bridge/clients/MessageClient.java
```diff
@@ -37,7 +37,7 @@
 
     T convertMessage(Message message);
 
-    void start();
+    void start() throws MessageClientException;
```
MikeDombo, line 40: so if start fails, what are we doing? Retry with backoff?

src/main/java/com/aws/greengrass/mqtt/bridge/clients/MQTTClient.java
jcosentino11: fixed

src/main/java/com/aws/greengrass/mqtt/bridge/clients/MessageClient.java
jcosentino11: looking more closely MQTTClient won't actually throw on start.....method had throws on it.   going to do some more cleanup

MikeDombo: what are the reasons why start can fail then?

If bridge can't connect to the local broker we shouldn't be erroring. We relatively commonly see in our UATs that the bridge takes a few attempts to connect.

MikeDombo: There's a keystoreexception which can't actually be thrown....

jcosentino11: connect happens async, and if exception happens yeah it retries w/ backoff

MikeDombo (approved):

jcosentino11:
seems like recent test failure was timeout in `KeystoreTest` on windows.  created https://github.com/aws-greengrass/aws-greengrass-mqtt-bridge/pull/99 to address

sameerzuberi (approved):

PR #101, opened 2023-03-17
feat: per-route mqtt5 config

jcosentino11:
**Issue #, if available:**

**Description of changes:**

Adds mqtt5-specific route options via a new `mqtt5RouteOptions` config.  This config is a mapping from route name to properties, similar to the existing `mqttTopicMapping` config option.
```
"mqtt5RouteOptions": {
    "route1": {
        "noLocal": false,
        "retainAsPublished": true
    },
    "route2": {
        ...
    },
    ...
}
```

Also configures objectmapper to allow unrecognized properties, so mqtt5 route config can grow over time.

⚠️ As discovered through this PR, we must not add fields to `mqttTopicMappings` as this will break backwards compatibility, due to objectmapper for current/previous bridge versions not being configured to allow unrecognized values.

### DIFF
--- a/src/main/java/com/aws/greengrass/mqtt/bridge/BridgeConfig.java
+++ b/src/main/java/com/aws/greengrass/mqtt/bridge/BridgeConfig.java
@@ -9,10 +9,12 @@ import com.aws.greengrass.config.Topics;
 import com.aws.greengrass.logging.api.Logger;
 import com.aws.greengrass.logging.impl.LogManager;
 import com.aws.greengrass.mqtt.bridge.model.InvalidConfigurationException;
+import com.aws.greengrass.mqtt.bridge.model.Mqtt5RouteOptions;
 import com.aws.greengrass.mqtt.bridge.model.MqttVersion;
 import com.aws.greengrass.util.Coerce;
 import com.aws.greengrass.util.Utils;
 import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.DeserializationFeature;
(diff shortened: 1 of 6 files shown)


MikeDombo (commented):
We _cannot_ break compatibility.


Let's talk about what else we can do to achieve our goal

jcosentino11:
Revised the PR to maintain compatibility as discussed

src/main/java/com/aws/greengrass/mqtt/bridge/BridgeConfig.java
```diff
@@ -32,19 +34,22 @@
...
 @EqualsAndHashCode
-@Builder
+@Builder(toBuilder = true)
 @RequiredArgsConstructor
 public final class BridgeConfig {
     private static final Logger LOGGER = LogManager.getLogger(BridgeConfig.class);
-
-    private static final JsonMapper OBJECT_MAPPER =
-            JsonMapper.builder().enable(MapperFeature.ACCEPT_CASE_INSENSITIVE_PROPERTIES).build();
+    private static final JsonMapper OBJECT_MAPPER = JsonMapper.builder()
```
MikeDombo, line 41: use our `SerializerFactory.getConfigurationSerializerJson()`

jcosentino11: fixed

MikeDombo (approved):

robcmann (approved):

PR #102, opened 2023-03-20
chore: remove global noLocal config, in favor of per-route

jcosentino11:
**Issue #, if available:**

**Description of changes:**
Removing the global `noLocal` config since we are using per-route `noLocal` instead.


jcosentino11:
> Not adding it in the right place yet?

It's defined in per-route config, but it's not being utilized yet

jcosentino11:
https://github.com/aws-greengrass/aws-greengrass-mqtt-bridge/blob/79f4dab397c2eadc88119d371ecc0b853528bde9/src/main/java/com/aws/greengrass/mqtt/bridge/model/Mqtt5RouteOptions.java#L19

MikeDombo (approved):

robcmann (approved):

PR #103, opened 2023-03-20
feat: async IotCoreClient

jcosentino11:
**Issue #, if available:**

**Description of changes:**

Updates `IotCoreClient` to use v5 spooler models, and adapts to CompletableFuture paradigm.

Since moving to CompletableFuture was significant work, took the opportunity to group subscription logic in a `SubscriptionManager` class, which with some modification could be reused for local v5 client.

There are some subtle behavior changes:
* if we lose mqtt connection, we'll stop our attempts to subscribe/unsubscribe and prevent future attempts until we reconnect
* this "circuit breaker" behavior also gets triggered on each individual connect/unsubscribe/publish if we detect that the mqtt client is offline
* as we shutdown, we prevent any further subscription attempts

There are also plenty of follow-up tasks which can be done in separate PRs, this change is to lay the groundwork

unit testing.  will need extensive manual and integration testing.

### DIFF
--- a/src/main/java/com/aws/greengrass/mqtt/bridge/clients/IoTCoreClient.java
+++ b/src/main/java/com/aws/greengrass/mqtt/bridge/clients/IoTCoreClient.java
@@ -251,6 +251,7 @@ public class IoTCoreClient implements MessageClient<com.aws.greengrass.mqtt.brid
                         }
                         return subscribeResult;
                     }, executorService)
+                    // TODO backoff
                     .exceptionally(e -> {
                         if (canSubscribe()) {
                             return subscribeWithRetry(topic).join(); // TODO is this join ok?
--- a/src/main/java/com/aws/greengrass/mqtt/bridge/clients/IoTCoreClient.java
+++ b/src/main/java/com/aws/greengrass/mqtt/bridge/clients/IoTCoreClient.java
@@ -9,56 +9,52 @@ import com.aws.greengrass.logging.api.Logger;
 import com.aws.greengrass.logging.impl.LogManager;
 import com.aws.greengrass.mqtt.bridge.model.Message;
 import com.aws.greengrass.mqttclient.MqttClient;
+import com.aws.greengrass.mqttclient.MqttRequestException;
 import com.aws.greengrass.mqttclient.PublishRequest;
-import com.aws.greengrass.mqttclient.SubscribeRequest;
-import com.aws.greengrass.mqttclient.UnsubscribeRequest;
-import com.aws.greengrass.util.RetryUtils;
+import com.aws.greengrass.mqttclient.v5.Publish;
+import com.aws.greengrass.mqttclient.v5.QOS;
+import com.aws.greengrass.mqttclient.v5.Subscribe;
+import com.aws.greengrass.mqttclient.v5.SubscribeResponse;
+import com.aws.greengrass.mqttclient.v5.Unsubscribe;
+import com.aws.greengrass.util.Utils;
 import lombok.AccessLevel;
 import lombok.Getter;
 import lombok.NonNull;
 import software.amazon.awssdk.crt.mqtt.MqttClientConnectionEvents;
-import software.amazon.awssdk.crt.mqtt.MqttMessage;
 import software.amazon.awssdk.crt.mqtt.QualityOfService;
 
-import java.time.Duration;
-import java.util.Arrays;
+import java.util.Collection;
 import java.util.HashSet;
 import java.util.Set;
-import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Consumer;
 import javax.inject.Inject;
 
 public class IoTCoreClient implements MessageClient<com.aws.greengrass.mqtt.bridge.model.MqttMessage> {
     private static final Logger LOGGER = LogManager.getLogger(IoTCoreClient.class);
-    public static final String TOPIC = "topic";
-    private final RetryUtils.RetryConfig subscribeRetryConfig =
-            RetryUtils.RetryConfig.builder().initialRetryInterval(Duration.ofSeconds(1L))
-                    .maxRetryInterval(Duration.ofSeconds(120L)).maxAttempt(Integer.MAX_VALUE)
-                    .retryableExceptions(Arrays.asList(ExecutionException.class, TimeoutException.class)).build();
+    private static final String LOG_KEY_TOPIC = "topic";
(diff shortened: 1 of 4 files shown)


src/main/java/com/aws/greengrass/mqtt/bridge/clients/IoTCoreClient.java
```diff
@@ -9,76 +9,66 @@
...
-                if (subscribeFuture != null) {
-                    subscribeFuture.cancel(true);
-                }
-                // subscribe to any topics left to be subscribed
-                Set<String> topicsToSubscribe = new HashSet<>(toSubscribeIotCoreTopics);
-                topicsToSubscribe.removeAll(subscribedIotCoreTopics);
-                subscribeFuture = executorService.submit(() -> subscribeToTopicsWithRetry(topicsToSubscribe));
-            }
+            subscriptionManager.online();
+            subscriptionManager.subscribeAll();
```
MikeDombo, line 71: note that if we've already subscribed successfully the spooler will handle the resubscription (if needed). If we haven't subscribed successfully yet, then yes you need to retry.

src/main/java/com/aws/greengrass/mqtt/bridge/clients/IoTCoreClient.java
```diff
@@ -108,32 +98,20 @@ public void start() {
...
-
-        this.subscribedIotCoreTopics.forEach(s -> {
-            try {
-                unsubscribeFromIotCore(s);
-                LOGGER.atDebug().kv(TOPIC, s).log("Unsubscribed from topic");
-            } catch (InterruptedException | ExecutionException | TimeoutException e) {
-                LOGGER.atWarn().kv(TOPIC, s).setCause(e).log("Unable to unsubscribe");
-            }
-        });
+        subscriptionManager.shutdown(1L, TimeUnit.MINUTES);
```
MikeDombo, line 101: don't block

src/main/java/com/aws/greengrass/mqtt/bridge/clients/IoTCoreClient.java
```diff
@@ -143,89 +121,260 @@ public void publish(com.aws.greengrass.mqtt.bridge.model.MqttMessage message) {
                 .payload(message.getPayload())
                 .qos(QualityOfService.AT_LEAST_ONCE)
                 .build();
-        iotMqttClient.publish(publishRequest);
+        // pubacks are not propagated
+        iotMqttClient.publish(publishRequest).exceptionally(e -> {
```
MikeDombo, line 125: use the updated publish method for mqtt5

src/main/java/com/aws/greengrass/mqtt/bridge/clients/IoTCoreClient.java
```diff
@@ -143,89 +121,260 @@ public void publish(com.aws.greengrass.mqtt.bridge.model.MqttMessage message) {
...
-                    // so that we can try to remove next time.
-                }
-            });
-
-            Set<String> topicsToSubscribe = new HashSet<>(topics);
-            topicsToSubscribe.removeAll(subscribedIotCoreTopics);
-
-            subscribeFuture = executorService.submit(() -> subscribeToTopicsWithRetry(topicsToSubscribe));
-        }
+        this.messageHandler = messageHandler;
```
MikeDombo, line 136: do we need to change the message handler all the time? That seems like it should only be set once.

src/main/java/com/aws/greengrass/mqtt/bridge/clients/IoTCoreClient.java
```diff
@@ -143,89 +121,260 @@ public void publish(com.aws.greengrass.mqtt.bridge.model.MqttMessage message) {
...
-            } catch (InterruptedException e) {
-                Thread.currentThread().interrupt();
-                return;
-            } catch (Exception e) {
-                LOGGER.atError().kv(TOPIC, topic).setCause(e).log("Failed to subscribe to IoTCore topic");
+                return iotMqttClient.subscribe(Subscribe.builder()
+                        .topic(topic)
+                        .callback(iotCoreCallback)
+                        // TODO support noLocal
+                        // TODO support retainAsPublished
```
MikeDombo, line 151: not needed

src/main/java/com/aws/greengrass/mqtt/bridge/clients/IoTCoreClient.java
```diff
@@ -143,89 +121,260 @@ public void publish(com.aws.greengrass.mqtt.bridge.model.MqttMessage message) {
...
+        /**
+         * Stop accepting new subscription requests, cancel all in-progress subscriptions,
+         * and unsubscribe.
+         */
+        public void shutdown(long timeout, TimeUnit unit) {
+            if (!shutdown.compareAndSet(false, true)) {
+                return;
+            }
+            try {
+                subscriptionManager.unsubscribeAll().get(timeout, unit);
```
MikeDombo, line 198: do not block

src/main/java/com/aws/greengrass/mqtt/bridge/clients/IoTCoreClient.java
```diff
@@ -143,89 +121,260 @@ public void publish(com.aws.greengrass.mqtt.bridge.model.MqttMessage message) {
...
+                topicsToSubscribe = new HashSet<>(toSubscribe);
+                topicsToSubscribe.removeAll(subscribed);
+            }
+            return subscribe(topicsToSubscribe);
+        }
+
+        public CompletableFuture<Void> subscribe(Collection<String> topics) {
+            CompletableFuture<Void> result = CompletableFuture.supplyAsync(() -> null, executorService);
+            for (String topic : topics) {
+                result.thenComposeAsync(unused -> {
```
MikeDombo, line 225: `CompletableFuture.allOf`?

src/main/java/com/aws/greengrass/mqtt/bridge/clients/IoTCoreClient.java
```diff
@@ -143,89 +121,260 @@ public void publish(com.aws.greengrass.mqtt.bridge.model.MqttMessage message) {
...
+                    } else {
+                        return CompletableFuture.completedFuture(null);
+                    }
+                }, executorService);
+            }
+            return result;
+        }
+
+        private CompletableFuture<Void> subscribeWithRetry(String topic) {
+            return IoTCoreClient.this.subscribe(topic).thenComposeAsync(resp -> {
```
MikeDombo, line 237: I don't see a reason why you need to compose async. Just composing should be fine. It will run on the MQTT client's thread, so be aware of this, but it should be fine?

MikeDombo (commented):
Can we split this? Use the MQTT 5 interfaces and treat them as blocking (use .get()). 
Then make it async, would that be reasonable or not worthwhile?

jcosentino11:
> Can we split this? Use the MQTT 5 interfaces and treat them as blocking (use .get()). Then make it async, would that be reasonable or not worthwhile?

That sounds reasonable---simpler impl to get off the ground and once we have solid battery of tests we can switch to async with more confidence

PR #107, opened 2023-03-23
chore: handle missing broker port in config

jcosentino11:
**Issue #, if available:**

**Description of changes:**

If no port is provided in brokerUri, default to 1883 or 8883, based on the brokerUri protocol being tcp or ssl

For upcoming LocalMqtt5Client which uses URI.getPort()

### DIFF
--- a/src/main/java/com/aws/greengrass/mqtt/bridge/BridgeConfig.java
+++ b/src/main/java/com/aws/greengrass/mqtt/bridge/BridgeConfig.java
@@ -103,11 +103,17 @@ public final class BridgeConfig {
                 configurationTopics.findOrDefault(DEFAULT_BROKER_URI, KEY_BROKER_SERVER_URI));
         String brokerUri = Coerce.toString(
                 configurationTopics.findOrDefault(brokerServerUri, KEY_BROKER_URI));
+        URI uri;
         try {
-            return new URI(brokerUri);
+            uri = new URI(brokerUri);
         } catch (URISyntaxException e) {
             throw new InvalidConfigurationException("Malformed " + KEY_BROKER_URI + ": " + brokerUri, e);
         }
+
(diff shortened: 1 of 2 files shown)


src/main/java/com/aws/greengrass/mqtt/bridge/BridgeConfig.java
```diff
@@ -103,11 +103,17 @@ private static URI getBrokerUri(Topics configurationTopics) throws InvalidConfig
...
                 configurationTopics.findOrDefault(brokerServerUri, KEY_BROKER_URI));
+        URI uri;
         try {
-            return new URI(brokerUri);
+            uri = new URI(brokerUri);
         } catch (URISyntaxException e) {
             throw new InvalidConfigurationException("Malformed " + KEY_BROKER_URI + ": " + brokerUri, e);
         }
+
+        if (uri.getPort() < 0) {
```
MikeDombo, line 113: We should have a default shouldn't we? Is this really what we did before your change?

jcosentino11: updated PR description, just checked paho docs,  if no port provided it'll use 1883/8883 as sensible, so not an existing bug

jcosentino11: will change to default to 1883/8883 based on the protocol

jcosentino11: so throwing an exception on missing port would break upgrade path.... will make sure not to do that

jcosentino11:
closing since config class doesn't seem like the right spot to handle this since, going to handle directly in LocalMqtt5Client instead

PR #110, opened 2023-03-27
chore: fix integration test flakiness

jcosentino11:
**Issue #, if available:**

**Description of changes:**

* Fix race condition in BridgeTest
* Refactor timeout values into constants

### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/BridgeTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/BridgeTest.java
@@ -110,8 +110,8 @@ public class BridgeTest {
         context.getIotCoreClient().getIotMqttClient().subscribe(Subscribe.builder()
                 .topic("topic/toIotCore")
                 .callback(p -> {
-                    messageReceived.countDown();
                     m.compareAndSet(null, MqttMessage.fromSpoolerV5Model(p));
+                    messageReceived.countDown();
                 })
                 .build());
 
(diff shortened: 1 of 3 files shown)


sameerzuberi (approved):

src/integrationtests/java/com/aws/greengrass/integrationtests/BridgeTest.java
```diff
@@ -90,7 +92,7 @@ void GIVEN_mqtt3_and_mapping_between_local_and_iotcore_WHEN_local_message_receiv
                         .contentType("contentType")
                         .build());
 
-        assertTrue(messageReceived.await(10L, TimeUnit.SECONDS));
+        assertTrue(messageReceived.await(AWAIT_TIMEOUT_SECONDS, TimeUnit.SECONDS));
```
MikeDombo, line 95: this is a _local_ message. Why would it take 30 seconds? This seems quite bad

jcosentino11: yeah, im still digging a bit more into why.  actually probably not a timeout, we're getting a null message on the assertion below

jcosentino11: think it was a race, latest commit has a fix

MikeDombo: you're counting down _before_ setting the value.

Use our asyncAssertOnConsumer, or use a future to get both a timeout and a value.

src/integrationtests/java/com/aws/greengrass/integrationtests/BridgeTest.java
```diff
@@ -28,6 +28,8 @@
 @BridgeIntegrationTest
 public class BridgeTest {
 
+    private static final long AWAIT_TIMEOUT_SECONDS = 30;
```
MikeDombo, line 31: bring this way down, should take like 2 sec?

jcosentino11: fixed

src/integrationtests/java/com/aws/greengrass/integrationtests/BridgeTest.java
jcosentino11: yeah, switched to asyncAssertOnConsumer 👍

MikeDombo (approved):

PR #112, opened 2023-03-28
refactor: injectable bridge config

jcosentino11:
**Issue #, if available:**

**Description of changes:**
Wrap `BridgeConfig` in an atomic reference and inject that wherever config is needed, for convenience.  `MQTTBridge` is responsible for updating the reference

Refactoring now in preparation for using route config within other parts of bridge.  It's easier to use bridge config if we can inject it, rather than using setters and making sure those are all properly called. (e.g. `localMqttClientFactory.setConfig(bridgeConfig);`)

### DIFF
--- a/src/main/java/com/aws/greengrass/mqtt/bridge/MQTTBridge.java
+++ b/src/main/java/com/aws/greengrass/mqtt/bridge/MQTTBridge.java
@@ -22,6 +22,7 @@ import com.aws.greengrass.mqtt.bridge.clients.LocalMqttClientFactory;
 import com.aws.greengrass.mqtt.bridge.clients.MessageClient;
 import com.aws.greengrass.mqtt.bridge.clients.MessageClientException;
 import com.aws.greengrass.mqtt.bridge.clients.PubSubClient;
+import com.aws.greengrass.mqtt.bridge.model.BridgeConfigReference;
 import com.aws.greengrass.mqtt.bridge.model.InvalidConfigurationException;
 import com.aws.greengrass.mqtt.bridge.model.MqttMessage;
 import com.aws.greengrass.mqttclient.MqttClient;
@@ -55,8 +56,8 @@ public class MQTTBridge extends PluginService {
     private MessageClient<MqttMessage> localMqttClient;
     private PubSubClient pubSubClient;
     private IoTCoreClient ioTCoreClient;
-    @Getter // for unit testing
-    private BridgeConfig bridgeConfig;
+    @Getter // for tests
+    private final BridgeConfigReference bridgeConfig;
 
 
     /**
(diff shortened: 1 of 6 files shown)


src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqttClientFactory.java
```diff
@@ -36,21 +43,22 @@ public LocalMqttClientFactory(MQTTClientKeyStore mqttClientKeyStore,
      * @throws MessageClientException if unable to create client
      */
     public MessageClient<MqttMessage> createLocalMqttClient() throws MessageClientException {
-        checkConfig();
+        BridgeConfig config = this.config.get();
+        if (config == null) {
+            throw new IllegalStateException("Bridge configuration not set");
```
MikeDombo, line 48: Illegal state shouldn't be used for this, it has a very very specific meaning.

We always prefer checked exceptions.

MikeDombo: Can it ever be null?

src/main/java/com/aws/greengrass/mqtt/bridge/MQTTBridge.java
```diff
@@ -243,30 +249,30 @@ public class ConfigurationChangeHandler {
...
             try {
-                bridgeConfig = BridgeConfig.fromTopics(configurationTopics);
+                newConfig = BridgeConfig.fromTopics(configurationTopics);
             } catch (InvalidConfigurationException e) {
                 serviceErrored(e);
                 return;
             }
 
-            localMqttClientFactory.setConfig(bridgeConfig);
+            BridgeConfig prevConfig = bridgeConfig.getAndUpdate(c -> newConfig);
```
MikeDombo, line 260: why not just getAndSet?

src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqttClientFactory.java
jcosentino11: i thought this was the specific meaning 😅   replaced w/ MessageClientException

MikeDombo: Nevermind that part, I removed it. I was thinking about a different exception type.

src/main/java/com/aws/greengrass/mqtt/bridge/MQTTBridge.java
jcosentino11: yep fixed

MikeDombo (approved):

sameerzuberi (approved):

src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqttClientFactory.java
jcosentino11: technically it can't be null, because we set the config during `install` and this factory is called within `startup`. more of a guard in case refactoring happens in the future

PR #113, opened 2023-03-28
feat: support noLocal

jcosentino11:
**Issue #, if available:**

**Description of changes:**

Support `noLocal` MQTT5 subscription property on `IotCoreClient` and `LocalMqtt5Client`.

To support MQTT5 features

TODO: unit and integration tests


src/main/java/com/aws/greengrass/mqtt/bridge/model/RouteLookup.java
```diff
@@ -0,0 +1,50 @@
...
+
+import com.aws.greengrass.mqtt.bridge.BridgeConfig;
+import com.aws.greengrass.mqtt.bridge.TopicMapping;
+import lombok.NonNull;
+
+import java.util.Map;
+import java.util.Optional;
+import javax.inject.Inject;
+
+public class RouteLookup {
```
MikeDombo, line 16: why do we need this class?

When we're building the routes we should know all their settings already and just pass that into the clients which need to subscribe/unsubscribe right?

jcosentino11: yeah that makes sense, original reason for the class was to contain the logic for cross-checking routes in topicMapping and mqtt5RouteOptions.  will refactor to store route stuff as a field in message clients

src/main/java/com/aws/greengrass/mqtt/bridge/BridgeConfig.java
```diff
@@ -197,6 +199,33 @@ private static long getSessionExpiryInterval(Topics configurationTopics) {
...
+     * @param source source type
+     * @return route options by topic name
+     */
+    public Map<String, Mqtt5RouteOptions> getMqtt5RouteOptionsForSource(TopicMapping.TopicType source) {
+        if (mqttVersion != MqttVersion.MQTT5) {
+            return Collections.emptyMap();
+        }
+        Map<String, Mqtt5RouteOptions> opts = new HashMap<>();
+        for (String route : topicMapping.keySet()) {
+            TopicMapping.MappingEntry mapping = topicMapping.get(route);
```
MikeDombo, line 217: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

**Problem**: Iterating over [map.keySet()](https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#keySet--) and calling [map.get(Key)](https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#get-java.lang.Object-) to get the values within the loop is inefficient. 
**Fix**: Iterate over [map.entrySet()](https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#entrySet--) and access the values using [entry.getValue()](https://docs.oracle.com/javase/8/docs/api/java/util/Map.Entry.html#getValue--) to avoid potential performance penalty due to lookup.

jcosentino11:
Error:    KeystoreTest.GIVEN_mqtt_bridge_WHEN_cda_ca_conf_changed_after_shutdown_THEN_bridge_keystore_not_updated:128 expected: <false> but was: <true>.   

Rerunning ci, will address this separately

src/main/java/com/aws/greengrass/mqtt/bridge/clients/IoTCoreClient.java
```diff
@@ -244,12 +250,16 @@ private void subscribeToTopicsWithRetry(Set<String> topics) {
 
     private void subscribeToIotCore(String topic)
```
MikeDombo, line 251: why isn't `nolocal` being passed in here? (or rather an object that represents the route)

jcosentino11: It can be, but we have that info in routesByTopic so we can just look it up without needing the param.  I don't feel strongly about it though 🤷

MikeDombo: Routes by topic is something you just added here right?

I'm seeing 19 files changed when it _feels_ like this should be 3 files plus a test

jcosentino11: Yeah it took some refactoring to be able to pass route config into IoTCoreClient because we need to read bridge config only once it's available.  This also touches both IoTCoreClient and LocalMqtt5Client, their tests and changes to integration test extension (because of aforementioned refactoring) Touches a lot of files but I think it's necessary. Could also split into multiple PRs

MikeDombo (approved):

src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5Client.java
```diff
@@ -362,8 +373,15 @@ private void subscribe(String topic) {
...
-                // TODO other mqtt5-specific fields
-                .withSubscription(topic, QOS.AT_LEAST_ONCE).build();
+                .withSubscription(
+                        topic,
+                        QOS.AT_LEAST_ONCE,
+                        isNoLocal(topic),
+                        // always set retainAsPublished so we have the retain flag available,
+                        // when we bridge messages, we'll set retain flag based on user route configuration.
+                        true,
+                        SubscribePacket.RetainHandlingType.SEND_ON_SUBSCRIBE_IF_NEW)
```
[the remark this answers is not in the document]
jcosentino11: no particular reason, fixed

MikeDombo (approved):

sameerzuberi (approved):

robcmann (approved):

PR #122, opened 2023-03-31
feat: handle retries on unsuback reason codes

sameerzuberi:
**Issue #, if available:**
Adding handling for UnSubAck reason codes. If the reason codes indicate a success or a non-retryable error, then it will not attempt to unsubscribe from the topic again. Otherwise the it will be retried according to the set RetryUtils configuration. 

Previously, no unsubscribe errors were being retried on. This change will retry on errors that have the possibility of succeeding on consecutive runs.

Unit tests

### DIFF
--- a/src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5Client.java
+++ b/src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5Client.java
@@ -437,8 +437,9 @@ public class LocalMqtt5Client implements MessageClient<MqttMessage> {
                         .log("Successfully subscribed to topic");
             } else if (subAckPacket.getReasonCodes().stream().allMatch(this::retrySubscribe)) {
                 // subscription failed with a retryable reason code, throw exception to trigger RetryUtils
-                throw new RetryableMqttOperationException("Failed to subscribe to " + topic + " with reason codes "
-                        + subAckPacket.getReasonCodes() + ": " + subAckPacket.getReasonString());
+                int rc = subAckPacket.getReasonCodes().stream().findFirst().get().getValue();
+                throw new RetryableMqttOperationException(String.format("Failed to subscribe to %s with reason code "
+                                + "%d: %s", topic, rc, subAckPacket.getReasonString()));
             } else {
                 // subscription failed with a non-retryable reason code
                 LOGGER.atError()
@@ -488,8 +489,9 @@ public class LocalMqtt5Client implements MessageClient<MqttMessage> {
                 LOGGER.atDebug().kv(LOG_KEY_TOPIC, topic).log("Unsubscribed from topic");
             } else if (unsubAckPacket.getReasonCodes().stream().allMatch(this::retryUnsubscribe)) {
                 // failed to unsubscribe with a retryable reason code, throw exception to trigger RetryUtils
-                throw new RetryableMqttOperationException("Failed to unsubscribe from " + topic + " with reason codes "
-                    + unsubAckPacket.getReasonCodes() + ": " + unsubAckPacket.getReasonString());
+                int rc = unsubAckPacket.getReasonCodes().stream().findFirst().get().getValue();
+                throw new RetryableMqttOperationException(String.format("Failed to unsubscribe from %s with reason "
+                        + "code %d: %s", topic, rc, unsubAckPacket.getReasonString()));
             } else {
                 // failed to unsubscribe with a non-retryable reason code
                 LOGGER.atDebug()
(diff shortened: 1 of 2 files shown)


src/test/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5ClientTest.java
```diff
@@ -340,6 +343,47 @@ void GIVEN_client_WHEN_update_subscriptions_with_null_message_handler_THEN_throw
...
+        topics.add(topic2);
+        doThrow(CrtRuntimeException.class).doNothing().when(clientSpy).unsubscribe(topic);
+
+        // subscribe to new topics, this will unsubscribe from the old topics
+        clientSpy.updateSubscriptions(topics, message -> {});
+        verify(clientSpy, times(2)).unsubscribe(topic);
+    }
+
+    @Test
+    void GIVEN_client_WHEN_unsubscribe_from_topic_with_retryable_reason_code_THEN_dont_retry() {
```
MikeDombo, line 370: retryable code, but not retrying?

src/test/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5ClientTest.java
```diff
@@ -340,6 +343,47 @@ void GIVEN_client_WHEN_update_subscriptions_with_null_message_handler_THEN_throw
...
+        Set<String> topics = new HashSet<>();
+        topics.add(topic);
+        LocalMqtt5Client clientSpy = spy(client);
+
+        // subscribe to topics
+        clientSpy.updateSubscriptions(topics, message -> {});
+
+        topics.remove(topic);
+        topics.add(topic2);
+        doThrow(CrtRuntimeException.class).doNothing().when(clientSpy).unsubscribe(topic);
```
MikeDombo, line 362: this is not showing that you retry on retryable reason codes. There is no reason code here, just an exception.

src/test/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5ClientTest.java
sameerzuberi: ah typo, will fix

src/test/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5ClientTest.java
sameerzuberi: I'll look further into getting a retryable reason code to trigger throw the exception itself. Looks like I'll need to set the Mqtt5 client's `subscribe` to return a subscribe packet with a reason code we're looking for

jcosentino11 (approved):

src/test/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5ClientTest.java
```diff
@@ -379,10 +386,35 @@ void GIVEN_client_WHEN_update_subscriptions_with_null_message_handler_THEN_throw
...
+        clientSpy.updateSubscriptions(topics, message -> {});
+
+        topics.remove(topic);
+        topics.add(topic2);
+        mockMqtt5Client.nextUnsubAckReasonCode.add(topic, UnsubAckPacket.UnsubAckReasonCode.UNSPECIFIED_ERROR);
+
+        // subscribe to new topics, this will unsubscribe from the old topics
+        clientSpy.updateSubscriptions(topics, message -> {});
+        assertThat("subscribed topics local client", clientSpy::getSubscribedLocalMqttTopics,
+                eventuallyEval(is(topics)));
```
MikeDombo, line 408: why eventually? why won't this be true immediately?

src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5Client.java
```diff
@@ -462,15 +483,22 @@ private void unsubscribe(String topic) {
         try {
             UnsubAckPacket unsubAckPacket = client.unsubscribe(unsubscribePacket)
                     .get(MQTT_OPERATION_TIMEOUT_MS, TimeUnit.MILLISECONDS);
-            if (!unsubAckPacket.getReasonCodes().contains(UnsubAckPacket.UnsubAckReasonCode.SUCCESS)) {
+            if (unsubAckPacket.getReasonCodes().stream().allMatch(this::unsubscribeIsSuccessful)) {
+                // successfully unsubscribed
+                LOGGER.atDebug().kv(LOG_KEY_TOPIC, topic).log("Unsubscribed from topic");
+            } else if (unsubAckPacket.getReasonCodes().stream().allMatch(this::retryUnsubscribe)) {
+                // failed to unsubscribe with a retryable reason code, throw exception to trigger RetryUtils
+                throw new RetryableMqttOperationException("Failed to unsubscribe from " + topic + " with reason codes "
```
MikeDombo, line 491: use string format

src/test/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5ClientTest.java
sameerzuberi: It is true immediately, the rest of the tests use eventually so I followed that, but I can update those as well

sameerzuberi: Just spoke with Joe about this, we're using eventually to future proof the tests in case we decide to make the clients fully async later on

MikeDombo (approved):

jcosentino11 (approved):

PR #123, opened 2023-03-31
feat: handle retries on suback reason codes

sameerzuberi:
**Description of changes:**
Adding handling for SubAck reason codes. If the reason codes indicate a success or a non-retryable error (such as not being authorized), then the subscription will not be retried. Otherwise the subscription will be retried according to the set RetryUtils configuration. 

Previously, no subscriptions errors were being retried on. This change will retry on errors that have the possibility of succeeding on consecutive runs.

Unit tests

### DIFF
--- a/src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5Client.java
+++ b/src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5Client.java
@@ -339,7 +339,7 @@ public class LocalMqtt5Client implements MessageClient<MqttMessage> {
     }
 
     @SuppressWarnings("PMD.AvoidCatchingGenericException")
-    void subscribeToTopics(Set<String> topics) {
+    private void subscribeToTopics(Set<String> topics) {
         for (String topic : topics) {
             try {
                 RetryUtils.runWithRetry(mqttExceptionRetryConfig, () -> {
@@ -389,6 +389,7 @@ public class LocalMqtt5Client implements MessageClient<MqttMessage> {
                         .kv(LOG_KEY_REASON, subAckPacket.getReasonString())
                         .kv(LOG_KEY_TOPIC, topic)
                         .log("Failed to subscribe to topic");
+                // throw exception to trigger RetryUtils
                 throw new CrtRuntimeException("Failed to subscribe to topic");
             }
         } catch (TimeoutException | ExecutionException e) {
--- a/src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5Client.java
+++ b/src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5Client.java
@@ -77,6 +77,8 @@ public class LocalMqtt5Client implements MessageClient<MqttMessage> {
     private static final String LOG_KEY_REASON = "reason";
     private static final String LOG_KEY_MESSAGE = "message";
     private static final String LOG_KEY_ERROR = "error";
+    public static final String RETRY = "retry";
+    public static final String NO_RETRY = "no retry";
 
     private static final long DEFAULT_TCP_MQTT_PORT = 1883;
     private static final long DEFAULT_SSL_MQTT_PORT = 8883;
@@ -110,6 +112,13 @@ public class LocalMqtt5Client implements MessageClient<MqttMessage> {
     private final Set<String> subscribedLocalMqttTopics = new HashSet<>();
     @Getter(AccessLevel.PACKAGE) // for testing
     private final Set<String> toSubscribeLocalMqttTopics = new HashSet<>();
+    private final Set<Integer> nonRetryableSubAckReasonCodes = new HashSet<Integer>(){{
+       this.add(SubAckPacket.SubAckReasonCode.NOT_AUTHORIZED.getValue());
+       this.add(SubAckPacket.SubAckReasonCode.TOPIC_FILTER_INVALID.getValue());
+       this.add(SubAckPacket.SubAckReasonCode.SHARED_SUBSCRIPTIONS_NOT_SUPPORTED.getValue());
+       this.add(SubAckPacket.SubAckReasonCode.SUBSCRIPTION_IDENTIFIERS_NOT_SUPPORTED.getValue());
+       this.add(SubAckPacket.SubAckReasonCode.WILDCARD_SUBSCRIPTIONS_NOT_SUPPORTED.getValue());
+    }};
     private Future<?> updateSubscriptionsTask;
 
     @Getter(AccessLevel.PACKAGE)
@@ -330,11 +339,11 @@ public class LocalMqtt5Client implements MessageClient<MqttMessage> {
     }
 
     @SuppressWarnings("PMD.AvoidCatchingGenericException")
-    private void subscribeToTopics(Set<String> topics) {
+    void subscribeToTopics(Set<String> topics) {
         for (String topic : topics) {
             try {
                 RetryUtils.runWithRetry(mqttExceptionRetryConfig, () -> {
-                    subscribe(topic); // TODO retry based on return code
+                    subscribe(topic);
                     return null;
                 }, "subscribe-mqtt5-topic", LOGGER);
             } catch (InterruptedException e) {
@@ -346,7 +355,7 @@ public class LocalMqtt5Client implements MessageClient<MqttMessage> {
         }
     }
 
-    private void subscribe(String topic) {
+    void subscribe(String topic) {
         Mqtt5Client client = getClient();
 
         if (!client.getIsConnected()) {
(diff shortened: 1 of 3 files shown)


src/test/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5ClientTest.java
```diff
@@ -340,6 +343,33 @@ void GIVEN_client_WHEN_update_subscriptions_with_null_message_handler_THEN_throw
...
+    @Test
+    void GIVEN_client_with_subscription_request_WHEN_retryable_reason_code_received_THEN_subscription_will_retry
+            (ExtensionContext context) {
+        ignoreExceptionOfType(context, CrtRuntimeException.class);
+        String topic = "iotcore/topic";
+        Set<String> topics = new HashSet<>();
+        topics.add(topic);
+        LocalMqtt5Client clientSpy = spy(client);
+
+        doThrow(CrtRuntimeException.class).doNothing().when(clientSpy).subscribe(topic);
```
MikeDombo, line 356: same problem: https://github.com/aws-greengrass/aws-greengrass-mqtt-bridge/pull/122/files#r1154883872

src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5Client.java
```diff
@@ -377,16 +387,24 @@ private void subscribe(String topic) {
...
+                        .kv(LOG_KEY_TOPIC, topic)
+                        .log("Failed to subscribe to topic with a non-retryable reason code, not retrying");
             } else {
                 LOGGER.atError()
                         .kv(LOG_KEY_REASON_CODES, subAckPacket.getReasonCodes())
                         .kv(LOG_KEY_REASON, subAckPacket.getReasonString())
                         .kv(LOG_KEY_TOPIC, topic)
                         .log("Failed to subscribe to topic");
+                // throw exception to trigger RetryUtils
+                throw new CrtRuntimeException("Failed to subscribe to topic");
```
jcosentino11, line 403: we should throw CrtRuntimeException within our code, should just be within crt library code.  either use an messageclientexception from bridge or maybe extend it e.g. RetryableMqttOperationException

src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5Client.java
```diff
@@ -87,6 +87,8 @@ public class LocalMqtt5Client implements MessageClient<MqttMessage> {
     private static final String LOG_KEY_REASON = "reason";
     private static final String LOG_KEY_MESSAGE = "message";
     private static final String LOG_KEY_ERROR = "error";
+    public static final String RETRY = "retry";
```
jcosentino11, line 90: unused?

src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5Client.java
```diff
@@ -118,6 +120,14 @@ public class LocalMqtt5Client implements MessageClient<MqttMessage> {
     private final Set<String> subscribedLocalMqttTopics = new HashSet<>();
     @Getter(AccessLevel.PACKAGE) // for testing
     private final Set<String> toSubscribeLocalMqttTopics = new HashSet<>();
+    @SuppressWarnings("PMD.DoubleBraceInitialization")
+    private final Set<Integer> nonRetryableSubAckReasonCodes = new HashSet<Integer>() {{
+       this.add(SubAckPacket.SubAckReasonCode.NOT_AUTHORIZED.getValue());
```
jcosentino11, line 125: can we add comments explaining why, similar to what spooler has

src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5Client.java
```diff
@@ -377,16 +387,24 @@ private void subscribe(String topic) {
                         .kv(LOG_KEY_REASON, subAckPacket.getReasonString())
                         .kv(LOG_KEY_TOPIC, topic)
                         .log("Successfully subscribed to topic");
+            } else if (subAckPacket.getReasonCodes().stream().allMatch(this::skipSubscribeRetry)) {
```
jcosentino11, line 390: i think the inverse would be easier to understand:
```
 if (any reason code is retryable) {
    retry
} else {
   just log
}
```

src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5Client.java
```diff
@@ -368,25 +386,38 @@ private void subscribe(String topic) {
...
             SubAckPacket subAckPacket = client.subscribe(subscribePacket)
                     .get(MQTT_OPERATION_TIMEOUT_MS, TimeUnit.MILLISECONDS);
-            if (subAckPacket.getReasonCodes().stream().allMatch(this::subscriptionIsSuccessful)) {
-                synchronized (subscriptionsLock) {
-                    subscribedLocalMqttTopics.add(topic);
-                }
-                LOGGER.atDebug()
+            if (subAckPacket.getReasonCodes().stream().allMatch(this::retrySubscribe)) {
+                // received a retryable reason code
+                LOGGER.atError()
```
jcosentino11, line 391: warn, since we're going to retry

src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5Client.java
```diff
@@ -368,25 +386,38 @@ private void subscribe(String topic) {
...
                         .kv(LOG_KEY_REASON_CODES, subAckPacket.getReasonCodes())
                         .kv(LOG_KEY_REASON, subAckPacket.getReasonString())
                         .kv(LOG_KEY_TOPIC, topic)
-                        .log("Successfully subscribed to topic");
+                        .log("Failed to subscribe to topic");
```
jcosentino11, line 395: also mention that we're retrying

src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5Client.java
```diff
@@ -368,25 +386,38 @@ private void subscribe(String topic) {
...
+                // throw exception to trigger RetryUtils
+                throw new RetryableMqttOperationException("Failed to subscribe to topic");
             } else {
+                if (subAckPacket.getReasonCodes().stream().allMatch(this::subscriptionIsSuccessful)) {
```
jcosentino11, line 399: nit: the ordering you had before was fine
```
if (subscription successful) {
   ...
} else if (retryable) {
   ...
} else {
   ...
}
```

sameerzuberi: Ah I mistook your other comment about the structure, I'll revert it

src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5Client.java
MikeDombo: Don't log anything here. But make the exception message logged by the retry utils useful enough to know what's going on. Include the topic and the reason cdoe.

MikeDombo: You don't change retryutils. The error message in the exception is what will be logged, so change that.

src/test/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5ClientTest.java
```diff
@@ -340,6 +341,35 @@ void GIVEN_client_WHEN_update_subscriptions_with_null_message_handler_THEN_throw
...
+        String topic = "iotcore/topic";
+        Set<String> topics = new HashSet<>();
+        topics.add(topic);
+        LocalMqtt5Client clientSpy = spy(client);
+
+        clientSpy.updateSubscriptions(topics, message -> {});
+
+        //verify that the subscription is not retried
+        verify(clientSpy, times(1)).subscribe(topic);
+    }
```
jcosentino11, line 372: and same here, verify the subscription state

src/test/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5ClientTest.java
```diff
@@ -340,6 +341,35 @@ void GIVEN_client_WHEN_update_subscriptions_with_null_message_handler_THEN_throw
...
+        ignoreExceptionOfType(context, RetryableMqttOperationException.class);
+        String topic = "iotcore/topic";
+        Set<String> topics = new HashSet<>();
+        topics.add(topic);
+
+        LocalMqtt5Client clientSpy = spy(client);
+        mockMqtt5Client.nextSubAckReasonCode.add(topic, SubAckPacket.SubAckReasonCode.UNSPECIFIED_ERROR);
+        clientSpy.updateSubscriptions(topics, message -> {});
+
+        verify(clientSpy, times(2)).subscribe(topic);
```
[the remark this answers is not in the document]
jcosentino11: and move ` verify(clientSpy, times(2)).subscribe(topic);` below the above ^ to prevent race condition

src/test/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5ClientTest.java
```diff
@@ -340,6 +341,35 @@ void GIVEN_client_WHEN_update_subscriptions_with_null_message_handler_THEN_throw
...
+    }
+
+    @Test
+    void GIVEN_client_with_subscription_request_WHEN_nonretryable_reason_code_received_THEN_dont_retry()
```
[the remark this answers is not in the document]
sameerzuberi: I think I can get rid of this test case and update the other test to verify that the subscribe method isn't retried

jcosentino11 (approved):

robcmann (approved):

PR #126, opened 2023-04-18
feat: use provided config values for withConnectOptions

sameerzuberi:
**Description of changes:**
Setting the `connectOptions` field of the CRT client builder based off of customer provided config values. 

Prior to this change the configurations set by the customer would not be applied to the client.

**How was this change tested?**
Wrote an integ test that reads in the new config values straight from the Bridge configuration file, then changes the config values and verifies they're updated in the client, and finally attempts to send a message that is too large and verifies the message is not published.


src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5Client.java
```diff
@@ -417,9 +439,7 @@ void subscribe(String topic) throws RetryableMqttOperationException {
                         topic,
                         QOS.AT_LEAST_ONCE,
                         isNoLocal(topic),
-                        // always set retainAsPublished so we have the retain flag available,
-                        // when we bridge messages, we'll set retain flag based on user route configuration.
-                        true,
+                        retainAsPublished(topic),
```
MikeDombo, line 442: Do not do this. Read the comment.

src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5Client.java
```diff
@@ -578,20 +604,22 @@ private Mqtt5Client createCrtClient() throws MessageClientException {
...
-                            Mqtt5ClientOptions.ClientOfflineQueueBehavior.FAIL_ALL_ON_DISCONNECT)
+                    .withOfflineQueueBehavior(Mqtt5ClientOptions.ClientOfflineQueueBehavior
+                            .FAIL_ALL_ON_DISCONNECT)
                     .withConnectOptions(new ConnectPacket.ConnectPacketBuilder()
                             .withRequestProblemInformation(true)
-                            .withClientId(clientId).build())
-                    // TODO configurable?
+                            .withClientId(clientId)
+                            .withSessionExpiryIntervalSeconds(sessionExpiryInterval)
+                            .withMaximumPacketSizeBytes(maximumPacketSize != null ? maximumPacketSize :
```
MikeDombo, line 618: withMaximumPacketSizeBytes can accept null which will use the default. Do not provide our own default.

MikeDombo (commented):
Test showing that setting these values actually works and that changing them in the configuration has an effect.

src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5Client.java
sameerzuberi: In that case do we need to read the retainAsPublished flag from the config in the local client at all? Might have misunderstood this description [https://sim.amazon.com/issues/GG-42541](https://sim.amazon.com/issues/GG-42541)

src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5Client.java
sameerzuberi: I received a null pointer error when I did that earlier, might've stemmed from something else so I'll revisit this

src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5Client.java
MikeDombo: No you do not need this flag to be read in the client.

MikeDombo: It needs to be used only when _routing_ the message to the destination (iot core)

jcosentino11 (approved):
assuming retainAsPublished will be separate PR?

src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5Client.java
```diff
@@ -108,6 +108,9 @@ public class LocalMqtt5Client implements MessageClient<MqttMessage> {
 
     private final URI brokerUri;
     private final String clientId;
+    public final long sessionExpiryInterval;
+    public final Long maximumPacketSize;
+    public final int receiveMaximum;
```
MikeDombo, line 113: don't make these public.

src/integrationtests/java/com/aws/greengrass/integrationtests/ConfigTest.java
```diff
@@ -78,6 +78,24 @@ void GIVEN_Greengrass_with_mqtt_bridge_WHEN_multiple_config_changes_consecutivel
         assertEquals(1, numRestarts.get());
     }
 
+    @TestWithMqtt5Broker
+    @WithKernel("mqtt5_connect_options.yaml")
+    void GIVEN_Greengrass_with_mqtt_bridge_WHEN_connect_options_set_in_config_THEN_local_client_uses_configured_values(Broker broker, ExtensionContext context) {
```
MikeDombo, line 83: prove that changing the config takes effect as well.

src/integrationtests/java/com/aws/greengrass/integrationtests/ConfigTest.java
```diff
@@ -78,6 +78,24 @@ void GIVEN_Greengrass_with_mqtt_bridge_WHEN_multiple_config_changes_consecutivel
...
+
+        // verify that the config is correctly read
+        assertEquals(expectedSessionExpiryInterval, testContext.getConfig().getSessionExpiryInterval());
+        assertEquals(expectedMaximumPacketSize, testContext.getConfig().getMaximumPacketSize());
+        assertEquals(expectedReceiveMaximum, testContext.getConfig().getReceiveMaximum());
+
+        // verify that the brokerClient config values are correctly set in the local client
+        assertEquals(expectedSessionExpiryInterval, testContext.getLocalV5Client().sessionExpiryInterval);
+        assertEquals(expectedMaximumPacketSize, testContext.getLocalV5Client().maximumPacketSize);
+        assertEquals(expectedReceiveMaximum, testContext.getLocalV5Client().receiveMaximum);
```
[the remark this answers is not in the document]
sameerzuberi: Got it, I'll update the test

src/integrationtests/java/com/aws/greengrass/integrationtests/ConfigTest.java
```diff
@@ -78,6 +89,88 @@ void GIVEN_Greengrass_with_mqtt_bridge_WHEN_multiple_config_changes_consecutivel
...
+                        .userProperties(Collections.singletonList(new UserProperty("key", "val")))
+                        .responseTopic("response topic")
+                        .messageExpiryIntervalSeconds(1234L)
+                        .payloadFormat(Publish.PayloadFormatIndicator.UTF8)
+                        .contentType("contentType")
+                        .build());
+
+        // We expect this to timeout since the local client should not have been able to publish the message due to
+        // its size, so the IoT core client will have no message to retrieve
+        assertThrows(TimeoutException.class,
```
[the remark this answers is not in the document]
sameerzuberi: Ah I see, I'll update it to use the local client to subscribe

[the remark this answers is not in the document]
sameerzuberi: Does [this test](https://github.com/aws-greengrass/aws-greengrass-mqtt-bridge/blob/main/src/integrationtests/java/com/aws/greengrass/integrationtests/BridgeTest.java#L135) need to be updated as well? Curious since it publishes and subscribes the same way I do here, are AWS credentials somehow set for that test?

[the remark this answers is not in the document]
sameerzuberi: Got it, makes sense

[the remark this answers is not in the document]
MikeDombo: and ensure that a small message is received correctly.

src/integrationtests/java/com/aws/greengrass/integrationtests/ConfigTest.java
```diff
@@ -78,6 +89,88 @@ void GIVEN_Greengrass_with_mqtt_bridge_WHEN_multiple_config_changes_consecutivel
...
+            }
+        });
+
+        config.lookup(BridgeConfig.KEY_SESSION_EXPIRY_INTERVAL).withValue(1);
+        config.lookup(BridgeConfig.KEY_MAXIMUM_PACKET_SIZE).withValue(1);
+        config.lookup(BridgeConfig.KEY_RECEIVE_MAXIMUM).withValue(1);
+
+        // verify that config changes take effect
+        assertTrue(bridgeRestarted.await(AWAIT_TIMEOUT_SECONDS, TimeUnit.SECONDS));
+        TimeUnit.MILLISECONDS.sleep(500);
```
jcosentino11, line 128: do we need this? instead could do eventuallyEquals below

src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5Client.java
```diff
@@ -151,6 +157,8 @@ public class LocalMqtt5Client implements MessageClient<MqttMessage> {
         this.add(SubAckPacket.SubAckReasonCode.WILDCARD_SUBSCRIPTIONS_NOT_SUPPORTED.getValue());
     }};
     private Future<?> updateSubscriptionsTask;
+    @Getter
+    private PublishPacket lastPublish;
```
MikeDombo, line 161: remove this. You do not need this to verify the functionality. You have a full broker and a full MQTT client running.

src/integrationtests/java/com/aws/greengrass/integrationtests/ConfigTest.java
```diff
@@ -78,6 +85,87 @@ void GIVEN_Greengrass_with_mqtt_bridge_WHEN_multiple_config_changes_consecutivel
...
+                .topic("topic/toLocal")
+                .payload("abc".getBytes(StandardCharsets.UTF_8))
+                .userProperties(Collections.singletonList(new UserProperty("key", "val")))
+                .responseTopic("response topic")
+                .messageExpiryIntervalSeconds(1234L)
+                .payloadFormat(Publish.PayloadFormatIndicator.UTF8)
+                .contentType("contentType")
+                .build();
+
+        testContext.getLocalV5Client().getClient().subscribe(new SubscribePacket.SubscribePacketBuilder()
```
MikeDombo, line 120: when you subscribe, there is code that is called for each message which is received on the subscription, that's what you need to validate.

Do not dive into the `getClient`, just use the public interface of the bridge client. Use `updateSubscriptions` to register a handler for your topic which will be called when there is a message.

src/integrationtests/java/com/aws/greengrass/integrationtests/ConfigTest.java
```diff
@@ -78,6 +87,90 @@ void GIVEN_Greengrass_with_mqtt_bridge_WHEN_multiple_config_changes_consecutivel
...
+                MqttMessage.builder()
+                        .topic(topic)
+                        .payload("abc".getBytes(StandardCharsets.UTF_8))
+                        .userProperties(Collections.singletonList(new UserProperty("key", "val")))
+                        .responseTopic("response topic")
+                        .messageExpiryIntervalSeconds(1234L)
+                        .payloadFormat(Publish.PayloadFormatIndicator.UTF8)
+                        .contentType("contentType")
+                        .build());
+        TimeUnit.MILLISECONDS.sleep(100);
```
jcosentino11, line 141: is there a way we can get rid of the sleep?

src/integrationtests/java/com/aws/greengrass/integrationtests/ConfigTest.java
```diff
@@ -78,6 +87,90 @@ void GIVEN_Greengrass_with_mqtt_bridge_WHEN_multiple_config_changes_consecutivel
...
+                .messageExpiryIntervalSeconds(1234L)
+                .payloadFormat(Publish.PayloadFormatIndicator.UTF8)
+                .contentType("contentType")
+                .build();
+
+        Consumer<MqttMessage> messageHandler = (message) -> {
+          assertEquals(Arrays.toString(expectedMessage.getPayload()), Arrays.toString(message.getPayload()));
+        };
+
+        testContext.getLocalV5Client().updateSubscriptions(topics, messageHandler);
```
jcosentino11, line 130: might be a good place for `asyncAssertOnConsumer`

src/integrationtests/java/com/aws/greengrass/integrationtests/ConfigTest.java
```diff
@@ -78,6 +87,90 @@ void GIVEN_Greengrass_with_mqtt_bridge_WHEN_multiple_config_changes_consecutivel
...
+                        .userProperties(Collections.singletonList(new UserProperty("key", "val")))
+                        .responseTopic("response topic")
+                        .messageExpiryIntervalSeconds(1234L)
+                        .payloadFormat(Publish.PayloadFormatIndicator.UTF8)
+                        .contentType("contentType")
+                        .build());
+        TimeUnit.MILLISECONDS.sleep(100);
+
+        // change config values
+        config.lookup(BridgeConfig.KEY_SESSION_EXPIRY_INTERVAL).withValue(1);
```
jcosentino11, line 144: lets wait until we receive the message first before changing config

src/integrationtests/java/com/aws/greengrass/integrationtests/ConfigTest.java
```diff
@@ -78,6 +87,90 @@ void GIVEN_Greengrass_with_mqtt_bridge_WHEN_multiple_config_changes_consecutivel
...
+                MqttMessage.builder()
+                        .topic(topic)
+                        .payload("this message is too large to be published".getBytes(StandardCharsets.UTF_8))
+                        .userProperties(Collections.singletonList(new UserProperty("key", "val")))
+                        .responseTopic("response topic")
+                        .messageExpiryIntervalSeconds(1234L)
+                        .payloadFormat(Publish.PayloadFormatIndicator.UTF8)
+                        .contentType("contentType")
+                        .build());
+        TimeUnit.MILLISECONDS.sleep(100);
```
jcosentino11, line 171: should also wait for the largeMEssageHandler to complete first so we know the message has been published

src/integrationtests/java/com/aws/greengrass/integrationtests/ConfigTest.java
sameerzuberi: Looking back I think I could try using a `Pair<CompletableFuture<Void>, Consumer<MqttMessage>` and wait on the completable future instead of sleeping to ensure the handler is triggered

src/integrationtests/java/com/aws/greengrass/integrationtests/ConfigTest.java
jcosentino11: i think i misspoke, we dont have a way to hook into that right now

src/integrationtests/java/com/aws/greengrass/integrationtests/ConfigTest.java
```diff
@@ -78,6 +93,92 @@ void GIVEN_Greengrass_with_mqtt_bridge_WHEN_multiple_config_changes_consecutivel
...
+
+        // verify that config changes take effect
+        assertThat("session expiry interval config update",
+                () -> testContext.getLocalV5Client().getSessionExpiryInterval(), eventuallyEval(is(1L)));
+        assertThat("maximum packet size config update", () -> testContext.getLocalV5Client().getMaximumPacketSize(),
+                eventuallyEval(is(1L)));
+        assertThat("receive maximum config update", () -> testContext.getLocalV5Client().getReceiveMaximum(),
+                eventuallyEval(is(1)));
+
+        // Publish a large message to IoT Core and verify that it is not received due to the local client's config
```
jcosentino11, line 162:
```suggestion
        // Publish a large message to the local broker and verify that it is not received due to the local client's config
```

src/integrationtests/java/com/aws/greengrass/integrationtests/ConfigTest.java
```diff
@@ -78,6 +93,92 @@ void GIVEN_Greengrass_with_mqtt_bridge_WHEN_multiple_config_changes_consecutivel
...
+        // We expect a timeout since the previous message was too large to successfully publish, given the local
```
jcosentino11, line 163: this and the above feel like theyre saying the same thing

sameerzuberi: missed that, updated the comments

jcosentino11 (approved):

MikeDombo (approved):

MikeDombo:
Thank you for the updates!

PR #129, opened 2023-04-28
feat: read retain as published from config when routing messages

sameerzuberi:
**Description of changes:**
Update the `MessageBridge` to read the `retainAsPublished` flag from the Mqtt5 route options set in the configuration. The flag will be set accordingly when the message is being routed to its destination. 

The `retainAsPublished` option was not being read in from the configuration previously.

Added a unit test which ensures that the retain flag is read properly from the Mqtt5 route options and is set accordingly.
Integration tests checking that configurations with and without `retainAsPublished` enabled route messages as expected.


src/main/java/com/aws/greengrass/mqtt/bridge/MessageBridge.java
```diff
@@ -146,10 +170,23 @@ private <T extends Message> void handleMessage(TopicMapping.TopicType sourceType
     private <T extends Message> void publishMessage(MessageClient<T> client, String topic, Message message)
             throws MessageClientException {
         T msg = client.convertMessage(message);
-        msg = (T) msg.newFromMessageWithTopic(topic);
+
+        if (isRetainAsPublished(topic)) {
+            // can only be true for MQTT messages
+            MqttMessage mqttMsg = (MqttMessage) msg;
```
jcosentino11, line 176: check type before casting

src/integrationtests/java/com/aws/greengrass/integrationtests/BridgeTest.java
```diff
@@ -166,4 +166,82 @@ void GIVEN_mqtt5_and_mapping_between_local_and_iotcore_WHEN_local_message_receiv
...
+                .userProperties(Collections.singletonList(new UserProperty("key", "val")))
+                .responseTopic("response topic")
+                .messageExpiryIntervalSeconds(1234L)
+                .payloadFormat(Publish.PayloadFormatIndicator.UTF8)
+                .contentType("contentType")
+                .retain(true)
+                .build();
+
+        Pair<CompletableFuture<Void>, Consumer<Publish>> subscribeCallback
+                = asyncAssertOnConsumer(p -> assertEquals(expectedMessage, MqttMessage.fromSpoolerV5Model(p)));
```
jcosentino11, line 187: can we subscribe/verify on local topic?  iot core doesn't support retain as published, so subscribing to iot core topic won't give us a message w/ retain value that we expect

jcosentino11 (commented):
can we also add unit tests?

src/main/java/com/aws/greengrass/mqtt/bridge/MessageBridge.java
```diff
@@ -146,8 +156,21 @@ private <T extends Message> void handleMessage(TopicMapping.TopicType sourceType
     private <T extends Message> void publishMessage(MessageClient<T> client, String topic, Message message)
             throws MessageClientException {
         T msg = client.convertMessage(message);
-        msg = (T) msg.newFromMessageWithTopic(topic);
-        client.publish(msg);
+
+        if (isRetainAsPublished(topic) && msg instanceof MqttMessage) {
+            MqttMessage mqttMsg = (MqttMessage) msg;
+            mqttMsg.setRetain(true);
```
MikeDombo, line 162: why are you setting retain? You need to read retain first so that it is set to the same value as it was published with, right?

src/main/java/com/aws/greengrass/mqtt/bridge/model/Mqtt5RouteOptions.java
```diff
@@ -17,7 +17,7 @@
 public class Mqtt5RouteOptions {
 
     public static final boolean DEFAULT_NO_LOCAL = false;
-    public static final boolean DEFAULT_RETAIN_AS_PUBLISHED = true;
+    public static final boolean DEFAULT_RETAIN_AS_PUBLISHED = false;
```
MikeDombo, line 20: why are you changing the default? Use the default that we decided in the design document.

src/main/java/com/aws/greengrass/mqtt/bridge/model/MqttMessage.java
```diff
@@ -19,6 +21,8 @@
 public class MqttMessage implements Message {
     String topic;
     byte[] payload;
+    @NonFinal
```
MikeDombo, line 24: everything is supposed to be final. To change a value use `toBuilder` and then `withRetain`

src/main/java/com/aws/greengrass/mqtt/bridge/model/Mqtt5RouteOptions.java
sameerzuberi: Discussed this with Joe, we think that the default was wrong the first time around. V3 doesn't support retain as published, so the default value should have been false to maintain backwards compatibility. I think the design doc may have been talking about keeping the flag true for the local client

src/main/java/com/aws/greengrass/mqtt/bridge/MessageBridge.java
sameerzuberi: I'm reading the retain setting from the mqtt5 route options in the `isRetainAsPublished` call, and then if the retain flag is set there then I update the message accordingly. I just updated this to use the builder instead of a setter

MikeDombo: The retain flag is not set by the route though.
The retain flag is in the incoming message.

sameerzuberi: Makes sense now, I had the wrong understanding of the purpose of the route. Now I'm setting the retain flag based off of the incoming message when the route has retain as published enabled

src/main/java/com/aws/greengrass/mqtt/bridge/MessageBridge.java
```diff
@@ -146,8 +156,21 @@ private <T extends Message> void handleMessage(TopicMapping.TopicType sourceType
     private <T extends Message> void publishMessage(MessageClient<T> client, String topic, Message message)
             throws MessageClientException {
         T msg = client.convertMessage(message);
-        msg = (T) msg.newFromMessageWithTopic(topic);
-        client.publish(msg);
+
+        if (isRetainAsPublished(topic) && msg instanceof MqttMessage) {
+            MqttMessage mqttMsg = (MqttMessage) msg;
+            mqttMsg = mqttMsg.toBuilder().retain(mqttMsg.isRetain()).build();
```
MikeDombo, line 162: you're reading from the same message that you're writing into. I don't think this is doing anything.

src/integrationtests/java/com/aws/greengrass/integrationtests/BridgeTest.java
```diff
@@ -166,4 +169,87 @@ void GIVEN_mqtt5_and_mapping_between_local_and_iotcore_WHEN_local_message_receiv
...
+                .responseTopic("response topic")
+                .messageExpiryIntervalSeconds(1234L)
+                .payloadFormat(Publish.PayloadFormatIndicator.UTF8)
+                .contentType("contentType")
+                .retain(true)
+                .build();
+
+        Pair<CompletableFuture<Void>, Consumer<MqttMessage>> messageHandler =
+                asyncAssertOnConsumer(message -> assertEquals(Arrays.toString(expectedMessage.getPayload()),
+                        Arrays.toString(message.getPayload())));
```
MikeDombo, line 195: you aren't validating retain

src/integrationtests/java/com/aws/greengrass/integrationtests/BridgeTest.java
```diff
@@ -166,4 +169,87 @@ void GIVEN_mqtt5_and_mapping_between_local_and_iotcore_WHEN_local_message_receiv
...
+                .responseTopic("response topic")
+                .messageExpiryIntervalSeconds(1234L)
+                .payloadFormat(Publish.PayloadFormatIndicator.UTF8)
+                .contentType("contentType")
+                .retain(false)
+                .build();
+
+        Pair<CompletableFuture<Void>, Consumer<MqttMessage>> messageHandler =
+                asyncAssertOnConsumer(message -> assertEquals(Arrays.toString(expectedMessage.getPayload()),
+                        Arrays.toString(message.getPayload())));
```
MikeDombo, line 237: you aren't validating retain

src/main/java/com/aws/greengrass/mqtt/bridge/MessageBridge.java
sameerzuberi: I'm updating the logic here to check if the route options have retain as published set to false. If they do, I can bridge the message without retain, otherwise I'll bridge it as is

src/main/java/com/aws/greengrass/mqtt/bridge/MessageBridge.java
```diff
@@ -146,8 +155,29 @@ private <T extends Message> void handleMessage(TopicMapping.TopicType sourceType
...
+    private boolean isRetainAsPublished(String topic) {
+        if (optionsByTopic.isEmpty()) {
+            return DEFAULT_RETAIN_AS_PUBLISHED;
+        }
+        Map<String, Mqtt5RouteOptions> opts = new HashMap<>();
+        for (Map.Entry<String, TopicMapping.MappingEntry> entry : topicMapping.getMapping().entrySet()) {
+            if (topic.equals(entry.getValue().getTopic())) {
+                String route = entry.getKey();
+                opts.put(entry.getValue().getTopic(), optionsByTopic.get(route));
+            }
```
MikeDombo, line 178: this is going to do a loop on every single publish. We really can't do that.

sameerzuberi: Makes sense, I'll try to find a way where we can pass in the optionsByTopic in the right format so we don't have to build it out each time

sameerzuberi: I edited it to pass in the optionsByTopic in the expected format when the message bridge is created at install

MikeDombo: Does that correctly handle updates when subscriptions and configurations change?

sameerzuberi: Hm as it stands it might not, I think I'll need to add the mqtt5 route options to the list of config changes that would require a reinstall

src/integrationtests/java/com/aws/greengrass/integrationtests/BridgeTest.java
```diff
@@ -166,4 +169,150 @@ void GIVEN_mqtt5_and_mapping_between_local_and_iotcore_WHEN_local_message_receiv
 
         subscribeCallback.getLeft().get(AWAIT_TIMEOUT_SECONDS, TimeUnit.SECONDS);
     }
+
+    @TestWithMqtt5Broker
+    @WithKernel("mqtt5_local_to_iotcore_retain.yaml")
```
jcosentino11, line 174: nit: can we name it ****_retain_as_published?  just to not cause confusion w/ basic retain

src/main/java/com/aws/greengrass/mqtt/bridge/BridgeConfig.java
```diff
@@ -323,6 +323,7 @@ public boolean reinstallRequired(BridgeConfig newConfig) {
                 || !Objects.equals(getMqttVersion(), newConfig.getMqttVersion())
                 || !Objects.equals(getSessionExpiryInterval(), newConfig.getSessionExpiryInterval())
                 || !Objects.equals(getMaximumPacketSize(), newConfig.getMaximumPacketSize())
-                || !Objects.equals(getReceiveMaximum(), newConfig.getReceiveMaximum());
+                || !Objects.equals(getReceiveMaximum(), newConfig.getReceiveMaximum())
+                || !Objects.equals(getMqtt5RouteOptions(), newConfig.getMqtt5RouteOptions());
```
jcosentino11, line 327: not for this PR, but we should think how to optimize this. technically this is only applicable to local mqtt source, but customers can still define route options for iot core source, even though it wouldn't have an effect..... that case would case reinstall, which we wouldn't want.

jcosentino11 (approved):

robcmann (approved):

PR #130, opened 2023-09-05
chore: remove client lock

jcosentino11:
**Issue #, if available:**

**Description of changes:**
Remove synchronization when setting internal mqtt client in LocalMqtt5Client, as this only happens during construction or while LocalMqtt5Client is stopped.


MikeDombo (commented):
Why is this safe? presumably this lock was put in place for a reason.

jcosentino11:
> Why is this safe? presumably this lock was put in place for a reason.

I think I was being too defensive when putting it in originally.  Client is being set only during a "reset" which happens when ca config changes.  This change is serialized because it's a config topic subscription, so this wouldn't overlap

MikeDombo (approved):

robcmann (approved):

PR #132, opened 2023-09-06
feat: local bridging perf test

jcosentino11:
**Issue #, if available:**

**Description of changes:**

Adds a performance test class which sends messages across the bridge, from to and from the local broker.

Performance tests are tagged in junit as `performance` and will be run using a new `performance` maven profile. This profile has been added to github actions as a separate job, as to not slow down the default unit/int testing job.

We can run these tests, in conjunction with profiling, to analyze the performance of bridge.


src/integrationtests/java/com/aws/greengrass/integrationtests/BridgePerformanceTest.java
```diff
@@ -0,0 +1,88 @@
...
+                failedPublishes.incrementAndGet();
+            }
+        }, 0L, (long) (TimeUnit.SECONDS.toMicros(1) / (double) tps), TimeUnit.MICROSECONDS);
+
+        testFinished.await();
+
+        ses.shutdown();
+        assertTrue(ses.awaitTermination(5L, TimeUnit.SECONDS));
+
+        assertThat("messages bridged", expectedPublishes::get, eventuallyEval(is(numMessagesBridged.get())));
```
MikeDombo, line 83: this doesn't really look right. If we expect the failed publishes to be 0, then you should be checking the number of messages bridged is exactly equal to a constant value (not an AtomicInteger)

jcosentino11:
tabling for now. thinking that it may be better to benchmark things like MessageBridge rather than testing a flow that goes through the broker

PR #134, opened 2023-09-07
chore: remove paho max inflight limit

jcosentino11:
**Issue #, if available:**

**Description of changes:**
Bridge MQTT3 client currently only allows 10 "inflight" QOS1 messages, which limits performance.  This change removes paho's inflight limit by setting it to max possible value.

### DIFF
--- a/src/main/java/com/aws/greengrass/mqtt/bridge/clients/MQTTClient.java
+++ b/src/main/java/com/aws/greengrass/mqtt/bridge/clients/MQTTClient.java
@@ -242,6 +242,7 @@ public class MQTTClient implements MessageClient<MqttMessage> {
     private MqttConnectOptions getConnectionOptions() throws KeyStoreException {
         MqttConnectOptions connOpts = new MqttConnectOptions();
         connOpts.setCleanSession(true);
+        connOpts.setMaxInflight(1000);
 
         if ("ssl".equalsIgnoreCase(brokerUri.getScheme())) {
             SSLSocketFactory ssf = mqttClientKeyStore.getSSLSocketFactory();


src/main/java/com/aws/greengrass/mqtt/bridge/clients/MQTTClient.java
```diff
@@ -242,6 +242,7 @@ private void updateSubscriptionsInternal() {
     private MqttConnectOptions getConnectionOptions() throws KeyStoreException {
         MqttConnectOptions connOpts = new MqttConnectOptions();
         connOpts.setCleanSession(true);
+        connOpts.setMaxInflight(Integer.MAX_VALUE);
```
MikeDombo, line 245: do we really want this?

What is the current default value?

jcosentino11: default value is 10

MikeDombo: What is the downside to increasing it?

jcosentino11: i think we shouldn't restrict inflight here, let the customer flood the system as much as they want--if we want a mechanism for backpressure / rate limiting i think it should be done in MessageBridge.

this would also make mqtt3 consistent with mqtt5 client, where we dont have inflight or flow control configured

jcosentino11: downside to increasing would be memory usage since messages would be sitting around waiting for their acknowledgement

MikeDombo: Inflight is configured automatically in mqtt 5 based on what the broker tells us. Having no limit sounds bad to me; paho will need to keep a data structure to know what has been acknowledged or not leading to huge memory usage. Set this to a more rational value like 1000 maybe.

jcosentino11:
looks like it broke tests though, taking a look: 
```BridgeTest.GIVEN_mqtt3_and_mapping_between_local_and_iotcore_WHEN_iotcore_message_received_THEN_message_bridged_to_local(Broker)[1] timed out waiting for local mqtt client to connect```

src/main/java/com/aws/greengrass/mqtt/bridge/clients/MQTTClient.java
jcosentino11: adjusted

MikeDombo (approved):

vabhasin (approved):

PR #136, opened 2023-09-08
refactor(mqtt3): cleanup MQTTClient

jcosentino11:
**Issue #, if available:**

**Description of changes:**
Refactor methods, adjust naming. Add an integration test for MQTTClient 

### DIFF
--- a/src/main/java/com/aws/greengrass/mqtt/bridge/clients/MQTTClient.java
+++ b/src/main/java/com/aws/greengrass/mqtt/bridge/clients/MQTTClient.java
@@ -42,9 +42,23 @@ import javax.net.ssl.SSLSocketFactory;
 public class MQTTClient implements MessageClient<MqttMessage> {
     private static final Logger LOGGER = LogManager.getLogger(MQTTClient.class);
 
+    private static final RetryUtils.RetryConfig RETRY_CONFIG_MQTT_EXCEPTION =
+            RetryUtils.RetryConfig.builder()
+                    .initialRetryInterval(Duration.ofSeconds(1L))
+                    .maxRetryInterval(Duration.ofSeconds(120L))
+                    .maxAttempt(Integer.MAX_VALUE)
+                    .retryableExceptions(Collections.singletonList(MqttException.class))
+                    .build();
+    private static final RetryUtils.RetryConfig RETRY_CONFIG_CONNECT =
+            RetryUtils.RetryConfig.builder()
+                    .initialRetryInterval(Duration.ofSeconds(1L))
+                    .maxRetryInterval(Duration.ofSeconds(120L))
+                    .maxAttempt(Integer.MAX_VALUE)
+                    // always attempt to retry
+                    .retryableExceptions(Collections.singletonList(Exception.class))
+                    .build();
+
     public static final String TOPIC = "topic";
-    private static final int MIN_WAIT_RETRY_IN_SECONDS = 1;
-    private static final int MAX_WAIT_RETRY_IN_SECONDS = 120;
 
     @Getter // for testing
     @Setter
@@ -84,11 +98,6 @@ public class MQTTClient implements MessageClient<MqttMessage> {
     };
 
     private final MQTTClientKeyStore mqttClientKeyStore;
-
-    private final RetryUtils.RetryConfig mqttExceptionRetryConfig =
-            RetryUtils.RetryConfig.builder().initialRetryInterval(Duration.ofSeconds(1L))
-                    .maxRetryInterval(Duration.ofSeconds(120L)).maxAttempt(Integer.MAX_VALUE)
-                    .retryableExceptions(Collections.singletonList(MqttException.class)).build();
 
     private final MqttCallback mqttCallback = new MqttCallback() {
         @Override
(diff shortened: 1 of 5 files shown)


src/main/java/com/aws/greengrass/mqtt/bridge/clients/MQTTClient.java
```diff
@@ -252,84 +259,115 @@ private MqttConnectOptions getConnectionOptions() throws KeyStoreException {
...
-
-        connectFuture = executorService.submit(this::reconnectAndResubscribe);
+    private void connectAndSubscribeAsync() {
+        synchronized (connectLock) {
+            if (connectFuture != null) {
+                connectFuture.cancel(true);
+            }
+            connectFuture = executorService.submit(() -> {
+                try {
+                    connect();
```
MikeDombo, line 297: is this required? Doesn't the client automatically reconnect when needed?

src/main/java/com/aws/greengrass/mqtt/bridge/clients/MQTTClient.java
```diff
@@ -252,84 +259,115 @@ private MqttConnectOptions getConnectionOptions() throws KeyStoreException {
...
     }
 
-    private synchronized void doConnect() throws MqttException, KeyStoreException {
-        if (!mqttClientInternal.isConnected()) {
-            mqttClientInternal.connect(getConnectionOptions());
+    @SuppressWarnings({
+            "PMD.AvoidCatchingGenericException", "PMD.AvoidRethrowingException", "PMD.ExceptionAsFlowControl"})
+    private void connect() throws InterruptedException {
+        if (mqttClientInternal.isConnected()) {
             LOGGER.atInfo()
```
MikeDombo, line 310: too much is being logged at info

src/main/java/com/aws/greengrass/mqtt/bridge/clients/MQTTClient.java
jcosentino11: there's an option in paho for auto reconnect but we don't use that afaik

src/main/java/com/aws/greengrass/mqtt/bridge/clients/MQTTClient.java
```diff
@@ -252,84 +259,115 @@ private MqttConnectOptions getConnectionOptions() throws KeyStoreException {
...
+                return null;
+            }, "connect-mqtt-topic", LOGGER);
+            LOGGER.atInfo()
+                    .kv(BridgeConfig.KEY_BROKER_URI, brokerUri)
+                    .kv(BridgeConfig.KEY_CLIENT_ID, clientId)
+                    .log("Connected to broker");
+        } catch (InterruptedException e) {
+            throw e;
+        } catch (Exception e) {
+            // TODO considering erroring the component, or other recovery options
```
MikeDombo, line 344: retrying?

src/main/java/com/aws/greengrass/mqtt/bridge/clients/MQTTClient.java
jcosentino11: will audit the class and adjust levels

MikeDombo (commented):
Looks like we aren't using persistent sessions, is that right? We should be able to save a lot of time when reconnecting if the session was persisted

src/main/java/com/aws/greengrass/mqtt/bridge/clients/MQTTClient.java
MikeDombo: Why not? Should we?

src/main/java/com/aws/greengrass/mqtt/bridge/clients/MQTTClient.java
jcosentino11: yeah that's fair, i suppose doesn't hurt to retry everything

jcosentino11:
> Looks like we aren't using persistent sessions, is that right? We should be able to save a lot of time when reconnecting if the session was persisted

yeah, was just going on original impl that didn't use persistent sessions so that wasn't in scope for me.   i think we may also need to switch from inmemorypersistence if we do persisten sessions? could do this as a follow up

MikeDombo:
> > Looks like we aren't using persistent sessions, is that right? We should be able to save a lot of time when reconnecting if the session was persisted
> 
> yeah, was just going on original impl that didn't use persistent sessions so that wasn't in scope for me. i think we may also need to switch from inmemorypersistence if we do persistent sessions? could do this as a follow up

No, the session I'm talking about is in the broker itself, not the client.

jcosentino11:
> > > Looks like we aren't using persistent sessions, is that right? We should be able to save a lot of time when reconnecting if the session was persisted
> > 
> > 
> > yeah, was just going on original impl that didn't use persistent sessions so that wasn't in scope for me. i think we may also need to switch from inmemorypersistence if we do persistent sessions? could do this as a follow up
> 
> No, the session I'm talking about is in the broker itself, not the client.

yeah-- ok i'll give this another pass with the stuff you suggested, should simplify things alot

jcosentino11:
Since all the async changes were included in 2.3.0 release, turned this into a refactor PR.  Will explore using cleanSession=false in a separate PR, although it is nice to not have to unsubscribe on shutdown :)

src/main/java/com/aws/greengrass/mqtt/bridge/clients/MQTTClient.java
```diff
@@ -228,16 +239,19 @@ public void publish(MqttMessage message) throws MessageClientException {
...
         if (mqttClientInternal.isConnected()) {
-            updateSubscriptionsInternal();
+            updateSubscriptionsAsync();
         }
     }
 
     @SuppressWarnings("PMD.AvoidCatchingGenericException")
-    private void updateSubscriptionsInternal() {
+    private void updateSubscriptionsAsync() {
         synchronized (subscribeLock) {
```
MikeDombo, line 255: this doesn't look async. We're unsubscribing from all topics synchronously. This method also then doesn't subscribe to any topics?

src/main/java/com/aws/greengrass/mqtt/bridge/clients/MQTTClient.java
```diff
@@ -42,11 +42,27 @@
...
+                    .maxAttempt(Integer.MAX_VALUE)
+                    .retryableExceptions(Collections.singletonList(MqttException.class))
+                    .build();
+    private static final RetryUtils.RetryConfig RETRY_CONFIG_CONNECT =
+            RetryUtils.RetryConfig.builder()
+                    .initialRetryInterval(Duration.ofSeconds(1L))
+                    .maxRetryInterval(Duration.ofSeconds(120L))
+                    .maxAttempt(Integer.MAX_VALUE)
+                    // always attempt to retry
+                    .retryableExceptions(Collections.singletonList(Exception.class))
```
MikeDombo, line 58: are we going to end up retrying on interrupted?

PR #141, opened 2023-09-13
chore: default to mqtt5

jcosentino11:
**Issue #, if available:**

**Description of changes:**
Set the default `mqtt/version` to `mqtt5` to match spooler.


MikeDombo (changes requested):
No. Do not do this.

You will break existing customers which is what we must never do.

PR #142, opened 2023-09-13
fix(mqtt5) react to client certificate rotations

jcosentino11:
**Issue #, if available:**

**Description of changes:**
Reset `LocalMqtt5Client` after CDA generates a client certificates so that it may used in the MQTT connection.

Client certificates rotate weekly by default and bridge mqtt5 client needs a way to use the new certificates.

##### Integration tests:
* rotate client cert and verify that local client resets
* rotate ca cert, server cert, and client cert, and verify local client resets
* rotate client certs in rapid succession, to test concurrency
##### Unit tests:
* verify that resets will be retried on failure
##### Memory:
Rotated client certs in a loop for 40 minutes, memory looks pretty stable
image

Both integration tests use certificates that are used by both broker (HiveMQ) and client.  the broker will dynamically reload on server certs and trust store changes.

### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/KeystoreTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/KeystoreTest.java
@@ -62,9 +62,8 @@ public class KeystoreTest {
     @WithKernel("mqtt5_config_ssl.yaml")
     void GIVEN_mqtt5_bridge_WHEN_client_cert_changes_THEN_memory_does_not_leak(Broker broker) throws Exception {
         while (true) {
-            CompletableFuture<Void> numConnects = asyncAssertNumConnects(1);
             testContext.getCerts().rotateClientCert();
-            numConnects.get(AWAIT_TIMEOUT_SECONDS, TimeUnit.SECONDS);
+            Thread.sleep(5000L);
         }
     }
 
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/KeystoreTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/KeystoreTest.java
@@ -24,6 +24,7 @@ import com.aws.greengrass.mqtt.bridge.auth.MQTTClientKeyStore;
 import com.aws.greengrass.util.Pair;
 import lombok.Builder;
 import lombok.Setter;
+import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.extension.ExtensionContext;
 import software.amazon.awssdk.crt.mqtt5.Mqtt5Client;
 import software.amazon.awssdk.crt.mqtt5.Mqtt5ClientOptions;
@@ -55,6 +56,17 @@ import static org.junit.jupiter.api.Assertions.fail;
 public class KeystoreTest {
     private static final long AWAIT_TIMEOUT_SECONDS = 30L;
     BridgeIntegrationTestContext testContext;
+
+    @Disabled("use for manual memory leak testing")
+    @TestWithMqtt5Broker
+    @WithKernel("mqtt5_config_ssl.yaml")
+    void GIVEN_mqtt5_bridge_WHEN_client_cert_changes_THEN_memory_does_not_leak(Broker broker) throws Exception {
+        while (true) {
+            CompletableFuture<Void> numConnects = asyncAssertNumConnects(1);
+            testContext.getCerts().rotateClientCert();
+            numConnects.get(AWAIT_TIMEOUT_SECONDS, TimeUnit.SECONDS);
+        }
+    }
 
     @TestWithMqtt5Broker
     @WithKernel("mqtt5_config_ssl.yaml")
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/KeystoreTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/KeystoreTest.java
@@ -22,7 +22,16 @@ import com.aws.greengrass.mqtt.bridge.BridgeConfig;
 import com.aws.greengrass.mqtt.bridge.MQTTBridge;
 import com.aws.greengrass.mqtt.bridge.auth.MQTTClientKeyStore;
 import com.aws.greengrass.util.Pair;
+import lombok.Builder;
+import lombok.Setter;
 import org.junit.jupiter.api.extension.ExtensionContext;
+import software.amazon.awssdk.crt.mqtt5.Mqtt5Client;
+import software.amazon.awssdk.crt.mqtt5.Mqtt5ClientOptions;
+import software.amazon.awssdk.crt.mqtt5.OnAttemptingConnectReturn;
+import software.amazon.awssdk.crt.mqtt5.OnConnectionFailureReturn;
+import software.amazon.awssdk.crt.mqtt5.OnConnectionSuccessReturn;
+import software.amazon.awssdk.crt.mqtt5.OnDisconnectionReturn;
+import software.amazon.awssdk.crt.mqtt5.OnStoppedReturn;
 
 import java.time.Instant;
 import java.util.Collections;
@@ -31,6 +40,7 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
+import java.util.stream.IntStream;
 
 import static com.aws.greengrass.componentmanager.KernelConfigResolver.CONFIGURATION_CONFIG_KEY;
 import static com.aws.greengrass.lifecyclemanager.GreengrassService.RUNTIME_STORE_NAMESPACE_TOPIC;
@@ -39,19 +49,56 @@ import static com.aws.greengrass.testcommons.testutilities.ExceptionLogProtector
 import static com.aws.greengrass.testcommons.testutilities.TestUtils.asyncAssertOnConsumer;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.fail;
 
 @BridgeIntegrationTest
 public class KeystoreTest {
     private static final long AWAIT_TIMEOUT_SECONDS = 30L;
-
     BridgeIntegrationTestContext testContext;
+
+    @TestWithMqtt5Broker
+    @WithKernel("mqtt5_config_ssl.yaml")
+    void GIVEN_mqtt5_bridge_WHEN_client_cert_changes_THEN_local_client_restarts(Broker broker) throws Exception {
+        CompletableFuture<Void> numConnects = asyncAssertNumConnects(1);
+        testContext.getCerts().rotateClientCert();
+        testContext.getCerts().waitForBrokerToApplyStoreChanges();
+        numConnects.get(AWAIT_TIMEOUT_SECONDS, TimeUnit.SECONDS);
+        assertTrue(testContext.getLocalV5Client().getClient().getIsConnected());
+    }
+
+    @TestWithMqtt5Broker
+    @WithKernel("mqtt5_config_ssl.yaml")
+    void GIVEN_mqtt5_bridge_WHEN_ca_changes_THEN_local_client_restarts(Broker broker) throws Exception {
+        CompletableFuture<Void> numConnects = asyncAssertNumConnects(1);
+        testContext.getCerts().rotateCA();
+        testContext.getCerts().rotateServerCert();
+        testContext.getCerts().rotateClientCert();
+        testContext.getCerts().waitForBrokerToApplyStoreChanges();
+        numConnects.get(AWAIT_TIMEOUT_SECONDS, TimeUnit.SECONDS);
+        assertTrue(testContext.getLocalV5Client().getClient().getIsConnected());
+    }
+
+    @TestWithMqtt5Broker
+    @WithKernel("mqtt5_config_ssl.yaml")
+    @SuppressWarnings("PMD.AvoidCatchingGenericException")
+    void GIVEN_mqtt5_bridge_WHEN_client_cert_changes_a_lot_THEN_local_client_is_connected(Broker broker) throws Exception {
+        IntStream.range(0, 30).forEach(i -> {
+            try {
+                testContext.getCerts().rotateClientCert();
+            } catch (Exception e) {
+                fail(e);
+            }
+        });
+        testContext.getCerts().waitForBrokerToApplyStoreChanges();
+        assertTrue(testContext.getLocalV5Client().getClient().getIsConnected());
+    }
 
     @TestWithMqtt3Broker
     @WithKernel("config.yaml")
     void GIVEN_mqtt_bridge_WHEN_cda_ca_conf_changed_THEN_bridge_keystore_updated(Broker broker) throws Exception {
         Pair<CompletableFuture<Void>, Consumer<Void>> keystoreUpdated = asyncAssertOnConsumer(p -> {}, 1);
         MQTTClientKeyStore keyStore = testContext.getKernel().getContext().get(MQTTClientKeyStore.class);
-        keyStore.listenToCAUpdates(() -> keystoreUpdated.getRight().accept(null));
+        keyStore.listenToUpdates(() -> keystoreUpdated.getRight().accept(null));
 
         Topic certificateAuthoritiesTopic = testContext.getKernel().getConfig().lookup(
                 SERVICES_NAMESPACE_TOPIC,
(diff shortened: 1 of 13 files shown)


src/integrationtests/java/com/aws/greengrass/integrationtests/KeystoreTest.java
```diff
@@ -37,21 +44,52 @@
...
+    }
+
+    @TestWithMqtt5Broker
+    @WithKernel("mqtt5_config_ssl.yaml")
+    void GIVEN_mqtt_bridge_WHEN_client_cert_changes_THEN_local_client_restarts(Broker broker) throws Exception {
+        CountDownLatch clientReset = new CountDownLatch(1);
+        executor.submit(() -> {
+            assertThat("client disconnected", () -> testContext.getLocalV5Client().getClient().getIsConnected(), eventuallyEval(is(false)));
+            assertThat("client reconnected", () -> testContext.getLocalV5Client().getClient().getIsConnected(), eventuallyEval(is(true)));
+            clientReset.countDown();
```
MikeDombo, line 72: this is going to be impossible to debug.

Use the test utilities async assert functionality.

src/main/java/com/aws/greengrass/mqtt/bridge/auth/MQTTClientKeyStore.java
```diff
@@ -56,6 +56,9 @@ public class MQTTClientKeyStore {
     @FunctionalInterface
     public interface UpdateListener {
         void onCAUpdate();
+
+        default void onClientCertUpdate() {
```
MikeDombo, line 60: why is this new? We already respond to cert rotations correctly in the mqtt 3 client, so why do we need this new callback, surely we must be handling it already

MikeDombo: Nevermind. I see

jcosentino11: for mqtt3, bridge just stores new cert in keystore, and mqtt3 client will just pick it up when it creates the new connection. it doesn't respond automatically to client cert updates currently

src/integrationtests/java/com/aws/greengrass/integrationtests/KeystoreTest.java
jcosentino11: 👍  will fix

src/main/java/com/aws/greengrass/mqtt/bridge/auth/MQTTClientKeyStore.java
```diff
@@ -135,16 +139,8 @@ public void updateCA(List<String> caCerts) throws IOException, CertificateExcept
             X509Certificate caCert = pemToX509Certificate(caCerts.get(i));
             keyStore.setCertificateEntry("CA" + i, caCert);
         }
-        setCaCerts(caCerts);
```
MikeDombo, line 138: why aren't we storing the ca cert anymore?

jcosentino11: should be there in the line below, just replaced setter w/ normal assignment

jcosentino11: hmm github may be reporting it weirdly

jcosentino11: yeah it's there it's weird because the line was part of setCaCerts

src/integrationtests/java/com/aws/greengrass/integrationtests/KeystoreTest.java
```diff
@@ -37,21 +44,52 @@
...
     BridgeIntegrationTestContext testContext;
 
+    @AfterEach
+    void tearDown() {
+        executor.shutdownNow();
+    }
+
+    @TestWithMqtt5Broker
+    @WithKernel("mqtt5_config_ssl.yaml")
+    void GIVEN_mqtt_bridge_WHEN_client_cert_changes_THEN_local_client_restarts(Broker broker) throws Exception {
```
MikeDombo, line 67: Does this test prove the client does not work as it is today (without the client cert listener)?

jcosentino11: yep, test fails if `reset` is removed from the client

src/integrationtests/java/com/aws/greengrass/integrationtests/KeystoreTest.java
jcosentino11: fixed

jcosentino11:
Ran the test in a loop for a few minutes, resources appear to be closed properly: 
image

src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5Client.java
```diff
@@ -589,31 +602,34 @@ private boolean retryUnsubscribe(UnsubAckPacket.UnsubAckReasonCode rc) {
 
     @Override
     public void start()  {
-        if (clientStarted) {
-            return;
+        if (isSSL()) {
```
MikeDombo, line 605: is `start` going to be called again if the ssl setting changes?

src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5Client.java
```diff
@@ -589,31 +602,34 @@ private boolean retryUnsubscribe(UnsubAckPacket.UnsubAckReasonCode rc) {
...
     private void closeClient() {
+        cancelUpdateSubscriptionsTask();
+        doCloseClient();
+    }
+
+    private void doCloseClient() {
         try {
-            if (clientStarted) {
+            if (client.getIsConnected()) {
                 client.stop(null);
```
MikeDombo, line 626: nit: stop with a disconnect packet showing normal disconnect.

src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5Client.java
```diff
@@ -589,31 +602,34 @@ private boolean retryUnsubscribe(UnsubAckPacket.UnsubAckReasonCode rc) {
...
             } else {
                 client.close();
```
MikeDombo, line 628: why is this in the else? It needs to be closed no matter what, so just remove the else?

src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5Client.java
```diff
@@ -589,31 +602,34 @@ private boolean retryUnsubscribe(UnsubAckPacket.UnsubAckReasonCode rc) {
...
             }
         } catch (CrtRuntimeException e) {
-            LOGGER.atError().setCause(e).log("Failed to stop MQTT5 client");
-        } finally {
-            clientStarted = false;
+            LOGGER.atDebug().setCause(e).log("Failed to stop MQTT5 client");
```
MikeDombo, line 631: why is this debug? This sounds like a problem

src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5Client.java
```diff
@@ -687,26 +703,57 @@ private Mqtt5Client createCrtClient() throws MessageClientException {
...
+            try {
+                this.client = clientSupplier.apply();
+            } catch (MessageClientException e) {
+                LOGGER.atWarn().cause(e).log("Unable to create mqtt client, will retry");
+                scheduleResetRetry();
+                return;
+            }
+
+            try {
+                client.start();
```
MikeDombo, line 731: can start actually fail? start connects asynchronously.

src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5Client.java
```diff
@@ -687,26 +703,57 @@ private Mqtt5Client createCrtClient() throws MessageClientException {
...
+                scheduleResetRetry();
+            }
+        }
+    }
+
+    private void scheduleResetRetry() {
+        synchronized (resetLock) {
+            cancelResetRetry();
+            resetRetryTask = ses.schedule(this::reset, resetDelay.getSeconds(), TimeUnit.SECONDS);
+            resetDelay = resetDelay.plus(resetDelay); // exponential backoff
```
MikeDombo, line 744: resetDelay needs to be set back to default on success

src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5Client.java
jcosentino11: its method signature says it can throw CrtRuntimeException, so in theory yes, although i haven't dug into the crt code to see the failure modes

src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5Client.java
jcosentino11: stop invokes the Mqtt5ClientOptions.LifecycleEvents#onStopped handler, which calls close

jcosentino11: i think the reason is to give a chance for the disconnect packet to reach the broker

src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5Client.java
jcosentino11: right now if ssl settings change, a new LocalMqtt5Client instance will be created.  after this change, we can use reset() instead of reinstalling for those config changes, and we'd need to revisit this.  but  this change addresses an issue where we were unsubscribing and resubscribing to keystore updates during a reset

src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5Client.java
MikeDombo: Do you expect that this branch would ever happen?

src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5Client.java
jcosentino11: fixed

src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5Client.java
jcosentino11: fixed

src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5Client.java
jcosentino11: fixed

src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5Client.java
```diff
@@ -88,11 +92,18 @@ public class LocalMqtt5Client implements MessageClient<MqttMessage> {
     private static final long DEFAULT_TCP_MQTT_PORT = 1883;
     private static final long DEFAULT_SSL_MQTT_PORT = 8883;
 
-    private boolean clientStarted = false; // crt close is not idempotent
+    private final MQTTClientKeyStore.UpdateListener onKeyStoreUpdate = new MQTTClientKeyStore.UpdateListener() {
+        @Override
+        public void onCAUpdate() {
+            LOGGER.atInfo().log("New CA cert available, reconnecting client");
+            reset();
```
MikeDombo, line 99: instead of directly calling reset, how about just scheduling the reset? That way you don't block this callback thread (which CDA owns) and you can avoid multiple threads calling into reset

MikeDombo: same for the other callback too of course

jcosentino11: yeah that's a good point, will give it a go

src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5Client.java
jcosentino11: thinking through this one a bit more...

jcosentino11: by branch you mean this `client.close()` right? that could happen if reset() is called during a retry scenario

src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5Client.java
jcosentino11: done

src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5Client.java
MikeDombo: The `else` branch, yes.

src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5Client.java
```diff
@@ -735,29 +733,29 @@ public void reset() {
...
         }
     }
 
-    private void scheduleResetRetry() {
+    private void scheduleResetTask() {
         synchronized (resetLock) {
-            cancelResetRetry();
-            resetRetryTask = ses.schedule(this::reset, resetDelay.getSeconds(), TimeUnit.SECONDS);
+            cancelResetTask();
+            resetTask = ses.schedule(this::reset, resetDelay.getSeconds(), TimeUnit.SECONDS);
```
MikeDombo, line 747: Use millis and the default delay should be probably 100ms, just to give a bit of time to de-dup requests, but still being quick to reset.

src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5Client.java
```diff
@@ -687,26 +706,59 @@ private Mqtt5Client createCrtClient() throws MessageClientException {
...
-        } catch (MessageClientException e) {
-            // TODO recover
-            LOGGER.atError().cause(e).log("unable to start mqtt client during reset");
-            return;
+    /**
+     * Stop and start the mqtt client.
+     */
+    public void reset() {
+        synchronized (resetLock) {
+            closeClient();
```
MikeDombo, line 721: So what's going on with publish/subscribe requests while we're doing the reset?

jcosentino11: subscribes will get retried after connection success, but publishes will be dropped

jcosentino11: bridge doesn't spooler LocalMqtt by itself, im not sure if we had that with paho before

MikeDombo: not just straight up dropped though, I'd expect some gross exceptions to happen. Let's try and make that cleaner by checking if the client exists (nonnull) and is connected before attempting to publish. And log a nicer debug message to say why.

Add an info log in reset to say that messages will be dropped until the connection is back up so that we're clear to customers and ourselves about what is going to happen.

jcosentino11: we already check for connection before publish, but no logs, will add

MikeDombo: OK, that's fine. I said debug, but I think trace for that since it is on every single message.

src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5Client.java
```diff
@@ -686,27 +733,68 @@ private Mqtt5Client createCrtClient() throws MessageClientException {
...
-            // TODO recover
-            LOGGER.atError().cause(e).log("unable to start mqtt client during reset");
-            return;
+    /**
+     * Stop and start the mqtt client.
+     */
+    public void reset() {
+        synchronized (resetLock) {
+            LOGGER.atWarn().log("Beginning client reset. The client will be offline for a period of time,"
+                    + "during which messages will dropped");
```
MikeDombo, line 754: space after `,` or before `d`

jcosentino11: 🙃 fixed

MikeDombo (approved):

sameerzuberi (approved):

PR #144, opened 2023-09-14
feat: handle unrecoverable errors

jcosentino11:
**Issue #, if available:**

**Description of changes:**
If an unrecoverable error happens in bridge's mqtt5 client, bubble the error up to the component level

In bridge's mqtt5 client, if we perform a reset, due to client certs or CA certs rotating, there's a chance that creating the new CRT client will fail.  However, I'm unsure of the exact failure mode.  Presumably the only thing in greengrass' control is the certs it configures the CRT client with, that are taken from CDA.  This would fail if certs provided by CDA are malformed, which in theory would never happen.  So if CRT client creation throws an exception, we can either 1) retry (likely wasteful), or 2) log and do nothing (bad), 3) error bridge.  My thinking is that it would be worth erroring the component so that the customer can be notified and fix the issue, since it doesn't appear that recovery within bridge would happen.

### DIFF
--- a/src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5Client.java
+++ b/src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5Client.java
@@ -736,11 +736,11 @@ public class LocalMqtt5Client implements MessageClient<MqttMessage> {
      * Stop and start the client.
      */
     public void reset() {
-        if (fatalErrorHandler.fatalErrorOccurred()) {
-            LOGGER.atWarn().log("Skipping client reset due to recent fatal error");
-            return;
-        }
         synchronized (restartLock) {
+            if (fatalErrorHandler.fatalErrorOccurred()) {
+                LOGGER.atWarn().log("Skipping client reset due to recent fatal error");
+                return;
+            }
             long stopTimeoutSeconds = 10L;
             Future<?> previousRestartTask = restartTask;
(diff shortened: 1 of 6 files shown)


src/main/java/com/aws/greengrass/mqtt/bridge/clients/LocalMqtt5Client.java
```diff
@@ -726,6 +737,10 @@ private boolean isSSL() {
      */
     public void reset() {
         synchronized (restartLock) {
+            if (fatalErrorHandler.fatalErrorOccurred()) {
+                LOGGER.atWarn().log("Skipping client reset due to recent fatal error");
+                return;
+            }
```
MikeDombo, line 743: why skip? How do we get out of this mode?

MikeDombo (commented):
Can you talk about the impetus for this change? What is wrong or deficient in the current implementation that this is helping with?

MikeDombo:
Ultimately, using `serviceErrored` is bad. It will cause the bridge to become broken and not recover. We do need to retry forever, with some appropriate backoff.

jcosentino11:
> Ultimately, using `serviceErrored` is bad. It will cause the bridge to become broken and not recover. We do need to retry forever, with some appropriate backoff.

Yeah the only reason for proposing this was because it didn't seem like retrying would actually help with recovery so service error would surface the issue more easily.  But yeah retrying forever works too and is certainly simpler :P 

Will close this

PR #154, opened 2023-09-26
fix: ensure connect loop stops on client close

jcosentino11:
**Issue #, if available:**

**Description of changes:**
1) Always attempt to disconnect, so paho can close the client (paho won't close if connect is in progress).  
2) Explicitly bail from connect loop if paho client is closed.
3) also fixes a bug where subscription state isn't cleared before a `reset()`

### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/ConfigTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/ConfigTest.java
@@ -20,6 +20,7 @@ import com.aws.greengrass.lifecyclemanager.GreengrassService;
 import com.aws.greengrass.mqtt.bridge.BridgeConfig;
 import com.aws.greengrass.mqtt.bridge.MQTTBridge;
 import com.aws.greengrass.mqtt.bridge.TopicMapping;
+import com.aws.greengrass.mqtt.bridge.clients.MQTTClient;
 import com.aws.greengrass.mqtt.bridge.model.BridgeConfigReference;
 import com.aws.greengrass.mqtt.bridge.model.Mqtt5RouteOptions;
 import com.aws.greengrass.mqtt.bridge.model.MqttMessage;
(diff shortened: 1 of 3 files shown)


src/integrationtests/java/com/aws/greengrass/integrationtests/ConfigTest.java
```diff
@@ -386,4 +388,39 @@ void GIVEN_Greengrass_with_mqtt_bridge_WHEN_invalid_mqttTopicMapping_updated_THE
 
         assertTrue(bridgeErrored.await(AWAIT_TIMEOUT_SECONDS, TimeUnit.SECONDS));
     }
+
+    @SuppressWarnings("PMD.AvoidCatchingGenericException")
+    @TestWithMqtt5Broker
+    @WithKernel("config.yaml")
+    void GIVEN_bridge_WHEN_mqtt_version_toggled_THEN_clients_switched(Broker broker, ExtensionContext context) throws Exception {
```
MikeDombo, line 395: This test reproduced issues we see in UAT?

src/main/java/com/aws/greengrass/mqtt/bridge/clients/MQTTClient.java
```diff
@@ -169,6 +162,32 @@ public void start() throws MessageClientException {
         connectAndSubscribe();
     }
 
+    private void disconnect() {
+        // 0ms quiescence time, just send the disconnect packet immediately
+        disconnect(0);
```
MikeDombo, line 167: May consider to use disconnect forcefully.

MikeDombo (approved):

src/integrationtests/java/com/aws/greengrass/integrationtests/ConfigTest.java
jcosentino11: yeah i observed "already connected" in logs and verified those go away with this test

jcosentino11: and as discussed still unsure about the shutdown timeout

PR #162, opened 2023-11-02
fix: improve MQTTClient shutdown reliability

jcosentino11:
**Issue #, if available:**

**Description of changes:**

force disconnect on shutdown
* fixes issue where disconnect attempt fails due to client being connected (race condition between cancelling MQTTClient's async connect task and calling disconnect)

With integration test from https://github.com/aws-greengrass/aws-greengrass-mqtt-bridge/pull/163 that triggers the condition from 1)

### DIFF
--- a/src/main/java/com/aws/greengrass/mqtt/bridge/clients/MQTTClient.java
+++ b/src/main/java/com/aws/greengrass/mqtt/bridge/clients/MQTTClient.java
@@ -162,7 +162,7 @@ public class MQTTClient implements MessageClient<MqttMessage> {
         connectAndSubscribe();
     }
 
-    @SuppressWarnings({"PMD.AvoidCatchingGenericException", "PMD.AvoidCatchingNPE"})
+    @SuppressWarnings({"PMD.AvoidCatchingGenericException", "PMD.AvoidCatchingNPE", "PMD.CloseResource"})
     private void disconnectForcibly() {
         IMqttClient client = mqttClientInternal;
         if (client == null) {
--- a/src/main/java/com/aws/greengrass/mqtt/bridge/clients/MQTTClient.java
+++ b/src/main/java/com/aws/greengrass/mqtt/bridge/clients/MQTTClient.java
@@ -209,7 +209,9 @@ public class MQTTClient implements MessageClient<MqttMessage> {
         try {
             // ensure that client cannot reconnect again
             // if callbacks would trigger for whatever reason
-            client.setCallback(null);
+            if (client != null) {
+                client.setCallback(null);
+            }
         } catch (NullPointerException ignore) {
             // can happen if client is closed.
             // ignoring as it is not a real error
--- a/src/main/java/com/aws/greengrass/mqtt/bridge/clients/MQTTClient.java
+++ b/src/main/java/com/aws/greengrass/mqtt/bridge/clients/MQTTClient.java
@@ -162,12 +162,17 @@ public class MQTTClient implements MessageClient<MqttMessage> {
         connectAndSubscribe();
     }
 
-    private void disconnect() {
+    @SuppressWarnings({"PMD.AvoidCatchingGenericException", "PMD.AvoidCatchingNPE"})
+    private void disconnectForcibly() {
         try {
-            // 0ms quiescence time, don't wait for DISCONNECT
-            mqttClientInternal.disconnectForcibly(0, 1L);
+            long doNotQuiesce = 0L;
+            long doNotWaitForDisconnectPacket = 1L; // since 0L means no timeout
+            mqttClientInternal.disconnectForcibly(doNotQuiesce, doNotWaitForDisconnectPacket);
         } catch (MqttException e) {
             LOGGER.atWarn().cause(e).log("Unable to disconnect forcibly");
+        } catch (NullPointerException ignore) {
+            // can happen if disconnectForcibly is called after the client is closed.
+            // since we're already disconnected, this is not a real error
         }
     }
 
@@ -201,7 +206,7 @@ public class MQTTClient implements MessageClient<MqttMessage> {
         mqttClientKeyStore.unsubscribeFromUpdates(onKeyStoreUpdate);
 
         cancelConnectTask();
-        disconnect();
+        disconnectForcibly();
 
         try {
             dataStore.close();
--- a/src/main/java/com/aws/greengrass/mqtt/bridge/clients/MQTTClient.java
+++ b/src/main/java/com/aws/greengrass/mqtt/bridge/clients/MQTTClient.java
@@ -164,10 +164,14 @@ public class MQTTClient implements MessageClient<MqttMessage> {
 
     @SuppressWarnings({"PMD.AvoidCatchingGenericException", "PMD.AvoidCatchingNPE"})
     private void disconnectForcibly() {
+        IMqttClient client = mqttClientInternal;
+        if (client == null) {
+            return;
+        }
         try {
             long doNotQuiesce = 0L;
             long doNotWaitForDisconnectPacket = 1L; // since 0L means no timeout
-            mqttClientInternal.disconnectForcibly(doNotQuiesce, doNotWaitForDisconnectPacket);
+            client.disconnectForcibly(doNotQuiesce, doNotWaitForDisconnectPacket);
         } catch (MqttException e) {
             LOGGER.atWarn().cause(e).log("Unable to disconnect forcibly");
         } catch (NullPointerException ignore) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/aws/greengrass/mqtt/bridge/clients/MQTTClient.java
```diff
@@ -203,12 +203,25 @@ public void stop() {
             LOGGER.atDebug().setCause(e).log("Unable to close mqtt client datastore");
         }
 
+        IMqttClient client = null;
         try {
-            IMqttClient client = mqttClientInternal;
```
MikeDombo, line 207: try the disconnectForcibly option instead of doing a nice disconnect in our `disconnect` method.

jcosentino11: yep that worked well, skips a bunch of checks that we preventing disconnect and sent the disconnect packet right away

src/main/java/com/aws/greengrass/mqtt/bridge/clients/MQTTClient.java
```diff
@@ -163,8 +163,15 @@ public void start() throws MessageClientException {
     }
 
     private void disconnect() {
-        // 0ms quiescence time, just send the disconnect packet immediately
-        disconnect(0);
+        if (!mqttClientInternal.isConnected()) {
+            return;
+        }
+        try {
+            // 0ms quiescence time, don't wait for DISCONNECT
```
MikeDombo, line 170: What's the 1 for?

src/main/java/com/aws/greengrass/mqtt/bridge/clients/MQTTClient.java
```diff
@@ -163,8 +163,15 @@ public void start() throws MessageClientException {
     }
 
     private void disconnect() {
-        // 0ms quiescence time, just send the disconnect packet immediately
-        disconnect(0);
+        if (!mqttClientInternal.isConnected()) {
```
MikeDombo, line 166: Races with call to close? What if it isn't connected but it gets connected immediately after this?

src/main/java/com/aws/greengrass/mqtt/bridge/clients/MQTTClient.java
jcosentino11: <= 0 in paho means wait forever 🙃

src/main/java/com/aws/greengrass/mqtt/bridge/clients/MQTTClient.java
jcosentino11: ah yeah thanks for the catch

src/main/java/com/aws/greengrass/mqtt/bridge/clients/MQTTClient.java
jcosentino11: https://github.com/eclipse/paho.mqtt.java/blob/f4e0db802a4433645ef011e711646a09ec9fae89/org.eclipse.paho.client.mqttv3/src/main/java/org/eclipse/paho/client/mqttv3/internal/Token.java#L142-L143

jcosentino11: so was just picking the next closest thing to "don't wait for disconnect packet"

jcosentino11: `MqttClient` has a method that doesn't send out disconnect packet, although it's not in `IMqttClient`

src/main/java/com/aws/greengrass/mqtt/bridge/clients/MQTTClient.java
jcosentino11: ran into null pointers with this change.  happens when force disconnect is called after close.  wouldn't expect this to happen with https://github.com/aws-greengrass/aws-greengrass-mqtt-bridge/pull/163, but i added a catch clause for this anyway for completeness, as awful as it is :) 
```
 java.lang.NullPointerException
	at org.eclipse.paho.client.mqttv3.internal.ClientComms.shutdownConnection(ClientComms.java:390)
	at org.eclipse.paho.client.mqttv3.internal.ClientComms.disconnectForcibly(ClientComms.java:559)
	at org.eclipse.paho.client.mqttv3.internal.ClientComms.disconnectForcibly(ClientComms.java:528)
	at org.eclipse.paho.client.mqttv3.MqttAsyncClient.disconnectForcibly(MqttAsyncClient.java:857)
	at org.eclipse.paho.client.mqttv3.MqttClient.disconnectForcibly(MqttClient.java:381)
	at com.aws.greengrass.mqtt.bridge.clients.MQTTClient.disconnect(MQTTClient.java:168)
	at com.aws.greengrass.mqtt.bridge.clients.MQTTClient.stop(MQTTClient.java:204)
	at com.aws.greengrass.mqtt.bridge.MQTTBridge.shutdown(MQTTBridge.java:165)
```
causes this, but that's fine

src/main/java/com/aws/greengrass/mqtt/bridge/clients/MQTTClient.java
```diff
@@ -162,9 +162,18 @@ public void start() throws MessageClientException {
...
-        disconnect(0);
+    @SuppressWarnings({"PMD.AvoidCatchingGenericException", "PMD.AvoidCatchingNPE"})
+    private void disconnectForcibly() {
+        try {
+            long doNotQuiesce = 0L;
+            long doNotWaitForDisconnectPacket = 1L; // since 0L means no timeout
+            mqttClientInternal.disconnectForcibly(doNotQuiesce, doNotWaitForDisconnectPacket);
+        } catch (MqttException e) {
+            LOGGER.atWarn().cause(e).log("Unable to disconnect forcibly");
+        } catch (NullPointerException ignore) {
```
MikeDombo, line 173: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

**Problem**: An exception has been ignored. The catch-block hides information about the stack trace. Not logging or forwarding the stack trace of an exception can mask unexpected errors and complicated debugging.

**Fix**: Use a catch block to handle the exception and write a log message declaring that the exception happened. 
[Learn more.](https://www.ibm.com/support/pages/best-practice-catching-and-re-throwing-java-exceptions)

jcosentino11: https://github.com/aws-greengrass/aws-greengrass-mqtt-bridge/pull/162#discussion_r1380909172

src/main/java/com/aws/greengrass/mqtt/bridge/clients/MQTTClient.java
MikeDombo: does calling close on the client do a forceful disconnect? Can we just call close without calling disconnect during shutdown? Given that we're trying to do a force disconnect anyway.

src/main/java/com/aws/greengrass/mqtt/bridge/clients/MQTTClient.java
MikeDombo: I mean to explain it in the comment. comment says 0ms, but doesn't mention the 1

src/main/java/com/aws/greengrass/mqtt/bridge/clients/MQTTClient.java
jcosentino11: it's a little different in impl, even w/ a force close, it'll throw if the client is connected or connecting. and do nothing if disconnect is in progress

MikeDombo: but this force disconnect also fails if it is not connected or currently connecting? So basically we have no way to stop it while it is connecting? We must instead wait for it to complete the connection?

src/main/java/com/aws/greengrass/mqtt/bridge/clients/MQTTClient.java
jcosentino11: ahh yeah i added a comment

src/main/java/com/aws/greengrass/mqtt/bridge/clients/MQTTClient.java
jcosentino11: force disconnect does not have the same behavior, doesn't make those same checks.  the way to stop while connecting would be disconnectForcibly

MikeDombo: ok that's good. I see the code no longer does a check for if you're connected or not, so that seems fine then.

MikeDombo (approved):

robcmann (approved):

PR #163, opened 2023-11-02
fix: synchronize lifecycle methods

jcosentino11:
**Issue #, if available:**

**Description of changes:**
Synchronize lifecycle methods in MQTTBridge

There's a chance that multiple MQTTClients can be active at the same time if startup and shutdown methods overlap, which is possible if reinstall is requested during startup.  When this occurs, the two MQTTClients step on each other, causing a connect/disconnect loop between them.  This behavior has been seen recently in customer logs.

New integration test, plus changes from https://github.com/aws-greengrass/aws-greengrass-mqtt-bridge/pull/162.  Verified that the new test fails without the fixes and passes with the fixes

### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/LifecycleTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/LifecycleTest.java
@@ -59,7 +59,7 @@ import static com.github.grantwest.eventually.EventuallyLambdaMatcher.eventually
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.is;
 import static org.junit.jupiter.api.Assertions.assertFalse;
-import static org.junit.jupiter.api.Assertions.fail;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.mockito.Mockito.mock;
 
 // not using @BridgeIntegrationTest because we're using a custom MQTTBridge service
@@ -120,9 +120,7 @@ class LifecycleTest {
     private static void launchKernel(Kernel kernel, ExtensionContext extensionContext, String configFileName) {
         System.setProperty("aws.greengrass.scanSelfClasspath", "true");
         URL configFile = extensionContext.getRequiredTestClass().getResource(configFileName);
-        if (configFile == null) {
-            fail("Config file " + configFileName + " not found");
-        }
+        assertNotNull(configFile);
         kernel.parseArgs("-i", configFile.toString());
         kernel.launch();
     }
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/LifecycleTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/LifecycleTest.java
@@ -1,0 +1,227 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.aws.greengrass.integrationtests;
+
+import com.aws.greengrass.builtin.services.pubsub.PubSubIPCEventStreamAgent;
+import com.aws.greengrass.clientdevices.auth.CertificateManager;
+import com.aws.greengrass.clientdevices.auth.api.ClientDevicesAuthServiceApi;
+import com.aws.greengrass.config.Topics;
+import com.aws.greengrass.dependency.ImplementsService;
+import com.aws.greengrass.dependency.State;
+import com.aws.greengrass.lifecyclemanager.Kernel;
+import com.aws.greengrass.mqtt.bridge.MQTTBridge;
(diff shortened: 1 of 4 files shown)


src/main/java/com/aws/greengrass/mqtt/bridge/MQTTBridge.java
```diff
@@ -119,60 +123,66 @@ protected MQTTBridge(Topics topics, TopicMapping topicMapping, MessageBridge mes
 
     @Override
     public void install() {
```
MikeDombo, line 125: any reason not to synchronize the method instead of putting the method body in a lock?

src/main/java/com/aws/greengrass/mqtt/bridge/MQTTBridge.java
```diff
@@ -119,60 +123,66 @@ protected MQTTBridge(Topics topics, TopicMapping topicMapping, MessageBridge mes
...
+                reportState(State.RUNNING);
+            } catch (MessageClientException e) {
+                serviceErrored(e);
+            }
         }
     }
 
     @Override
     public void shutdown() {
-        certificateAuthorityChangeHandler.stop();
```
MikeDombo, line 160: seems like the main problem is that shutdown either isn't invoked or it isn't actually stopping the client fully thus allowing it to keep connecting.

src/main/java/com/aws/greengrass/mqtt/bridge/MQTTBridge.java
jcosentino11: no reason other than naming, can move it to the method

src/main/java/com/aws/greengrass/mqtt/bridge/MQTTBridge.java
jcosentino11: yeah, i've noticed in debugging that the client can be null during shutdown (even if client variable is volatile)

jcosentino11: so if the timing is unfortunate enough, shutdown happens just before client is created during startup, creating the leak

jcosentino11: found that checking for interrupt before creating/starting the client works too, but there's still a race involved technically

MikeDombo: how can that happen though? startup would be interrupted, so it should then shutdown

MikeDombo: OK, so startup is interrupted, but that isn't resulting in the client _not_ trying to connect?

jcosentino11: was thinking about this scenario
```
startup() {
   if (interrupted) {
       return;
   }
   // shutdown() could execute theoretically before the next line, causing a leak?
   client = createMQTTClient()
}
```

MikeDombo: yes, but creating the client should fail (possibly after attempting to connect) _or_ the client should not be re-created when starting again

jcosentino11: ah i see, will check that

src/integrationtests/java/com/aws/greengrass/integrationtests/LifecycleTest.java
```diff
@@ -0,0 +1,227 @@
...
+        kernel.getContext().put(LocalMqttClientFactory.class, kernel.getContext().get(CachingLocalMqttClientFactory.class));
+        return kernel;
+    }
+
+    private static void launchKernel(Kernel kernel, ExtensionContext extensionContext, String configFileName) {
+        System.setProperty("aws.greengrass.scanSelfClasspath", "true");
+        URL configFile = extensionContext.getRequiredTestClass().getResource(configFileName);
+        if (configFile == null) {
+            fail("Config file " + configFileName + " not found");
+        }
```
MikeDombo, line 125: assertNotNull is fine

MikeDombo (dismissed):

jcosentino11:
test should be failing here but it's not.  it fails fine on my local, so going to see if i can make the integ test more reliable

jcosentino11:
> test should be failing here but it's not. it fails fine on my local, so going to see if i can make the integ test more reliable

So thinking through this more, it fails on my local because the timing for MQTTClient shutdown/startup is close enough to trigger condition where client isn't fully closed.   

When removing synchronization the test does fail on github, https://github.com/aws-greengrass/aws-greengrass-mqtt-bridge/pull/164, so seems good.

robcmann (approved):

MikeDombo (approved):

